PR #13, opened 2016-04-28
Always create a tar.gz for compressed artifacts, since docker only automatically uncompresses tar files.

Page-:


automation/jenkins_build.sh
```diff
@@ -58,11 +58,11 @@ mkdir -p $BUILD_DEPLOY_DIR
...
 if [ "${COMPRESSED}" == 'true' ]; then
-	if [ "${ARCHIVE}" == 'true' ]; then
-		(cd $BUILD_DEPLOY_DIR && tar --remove-files -zcvf ${DEPLOY_ARTIFACT}.tar.gz $DEPLOY_ARTIFACT)
-	else
-		gzip $BUILD_DEPLOY_DIR/$DEPLOY_ARTIFACT
+	if [ "${ARCHIVE}" != 'true' ]; then
+		 mv $BUILD_DEPLOY_DIR/$DEPLOY_ARTIFACT $BUILD_DEPLOY_DIR/resin.img
+		 DEPLOY_ARTIFACT=resin.img
 	fi
+	(cd $BUILD_DEPLOY_DIR && tar --remove-files -zcvf ${DEPLOY_ARTIFACT}.tar.gz $DEPLOY_ARTIFACT)
```
petrosagg: You can use `tar --remove-files -cvO $DEPLOY_ARTIFACT  | pigz -9 > ${DEPLOY_ARTIFACT}.tar.gz` for parallel and higher rate compression

automation/jenkins_build.sh
```diff
@@ -58,11 +58,11 @@ mkdir -p $BUILD_DEPLOY_DIR
 rm -rf $BUILD_DEPLOY_DIR/* # do we have anything there?
 mv -v $(readlink --canonicalize $WORKSPACE/build/tmp/deploy/images/$MACHINE/$DEPLOY_ARTIFACT) $BUILD_DEPLOY_DIR/$DEPLOY_ARTIFACT
 if [ "${COMPRESSED}" == 'true' ]; then
-	if [ "${ARCHIVE}" == 'true' ]; then
-		(cd $BUILD_DEPLOY_DIR && tar --remove-files -zcvf ${DEPLOY_ARTIFACT}.tar.gz $DEPLOY_ARTIFACT)
-	else
-		gzip $BUILD_DEPLOY_DIR/$DEPLOY_ARTIFACT
+	if [ "${ARCHIVE}" != 'true' ]; then
```
agherzan: I don't quite understand where is the archiving in this if.

Page-: This is whether yocto generates an archive or not, ie a directory - if it's not a directory then we need to standardise the name to resin.img to play nicely with automatically extracting with the right filename (since it has to be a tar file to be automatically decompressed)

PR #17, opened 2016-05-24
Support for yocto builds in docker

agherzan:
@floion @telphan @Page- @petrosagg

### DIFF
--- a/automation/Dockerfile
+++ b/automation/Dockerfile
@@ -1,0 +1,21 @@
+FROM ubuntu:15.04
+
+# Install the following utilities (required by poky)
+RUN apt-get update && apt-get install -y build-essential chrpath curl diffstat gcc-multilib gawk git-core libsdl1.2-dev texinfo unzip wget xterm cpio file && rm -rf /var/lib/apt/lists/*
+
+# Additional host packages required by resin
+RUN apt-get update && apt-get install -y jq npm nodejs sudo && rm -rf /var/lib/apt/lists/*
+
+# Fix path of node/nodejs
+RUN ln -s /usr/bin/nodejs /usr/bin/node
+
+# Install docker
+# https://github.com/docker/docker/blob/master/project/PACKAGERS.md#runtime-dependencies
+RUN apt-get update && apt-get install -y iptables procps e2fsprogs xfsprogs xz-utils git kmod && rm -rf /var/lib/apt/lists/*
+ENV DOCKER_VERSION 1.10.1
+VOLUME /var/lib/docker
+RUN wget -q -O /usr/bin/docker https://get.docker.com/builds/Linux/x86_64/docker-${DOCKER_VERSION} && chmod +x /usr/bin/docker
+
+COPY prepare-and-start.sh /
+
+WORKDIR /yocto/resin-board
--- a/automation/Dockerfile
+++ b/automation/Dockerfile
@@ -13,8 +13,10 @@ RUN ln -s /usr/bin/nodejs /usr/bin/node
 # https://github.com/docker/docker/blob/master/project/PACKAGERS.md#runtime-dependencies
 RUN apt-get update && apt-get install -y iptables procps e2fsprogs xfsprogs xz-utils git kmod && rm -rf /var/lib/apt/lists/*
 ENV DOCKER_VERSION 1.10.1
+ENV DOCKER_SHA256 de4057057acd259ec38b5244a40d806993e2ca219e9869ace133fad0e09cedf2
 VOLUME /var/lib/docker
-RUN wget -q -O /usr/bin/docker https://get.docker.com/builds/Linux/x86_64/docker-${DOCKER_VERSION} && chmod +x /usr/bin/docker
+RUN wget -q -O /usr/bin/docker https://get.docker.com/builds/Linux/x86_64/docker-${DOCKER_VERSION} && chmod +x /usr/bin/docker \
+    && echo "${DOCKER_SHA256}" /usr/bin/docker | sha256sum -c -
 
 COPY prepare-and-start.sh /
 
--- a/automation/jenkins_build.sh
+++ b/automation/jenkins_build.sh
@@ -1,6 +1,22 @@
 #!/bin/bash
 set -e
 
+BUILD_CONTAINER_NAME=yocto-build-$$
+
+cleanup() {
+    echo "[INFO] jenkins_build.sh: Cleanup."
+
+    # Stop docker container
+    echo "[INFO] jenkins_build.sh: Cleaning up yocto-build container."
+    docker stop $BUILD_CONTAINER_NAME 2> /dev/null || true
+    docker rm --volumes $BUILD_CONTAINER_NAME 2> /dev/null || true
+
+    if [ "$1" == "fail" ]; then
+        exit 1
+    fi
+}
+trap 'cleanup fail' SIGINT SIGTERM
+
 MACHINE=$1
 JENKINS_PERSISTENT_WORKDIR=$2
 JENKINS_DL_DIR=$JENKINS_PERSISTENT_WORKDIR/shared-downloads
@@ -12,8 +28,8 @@ if [ "$#" -ne 2 ]; then
     echo "Usage: jenkins_build.sh <MACHINE> <JENKINS_PERSISTENT_WORKDIR>"
     exit 1
 fi
-if [ -z "$BUILD_NUMBER" ] || [ -z "$sourceBranch" ]; then
-    echo "[ERROR] BUILD_NUMBER and sourceBranch variable undefined."
+if [ -z "$BUILD_NUMBER" ] || [ -z "$WORKSPACE" ] || [ -z "$sourceBranch" ] || [ -z "$metaResinBranch" ] || [ -z "$supervisorTag" ]; then
+    echo "[ERROR] BUILD_NUMBER, WORKSPACE, sourceBranch, metaResinBranch and supervisorTag are required."
     exit 1
 fi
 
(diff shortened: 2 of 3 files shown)


automation/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM ubuntu:15.04
+
+# Update index
+RUN apt-get update && apt-get -y upgrade
```
Page-: It would probably be better to remove this and instead use `docker build --pull ...` to build the image, ie make sure we're using the latest version of the `ubuntu:15.04` image

automation/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+# Install the following utilities (required by poky)
+RUN apt-get install -y build-essential chrpath curl diffstat gcc-multilib gawk git-core libsdl1.2-dev texinfo unzip wget xterm cpio file
```
Page-: This needs to be `apt-get update && apt-get install ...` otherwise we may hit a case of trying to build with out of date package lists and having it fail due to that (same applies for all other `apt-get install`s)

Page-: This should end with `&& rm -rf /var/lib/apt/lists/*` in order to save some space in the final image (just makes downloading a bit quicker, and use less storage space)

automation/jenkins_build.sh
```diff
@@ -26,25 +27,36 @@ fi
...
-    --log \
-    --remove-build \
-    --machine "$MACHINE" \
-    --supervisor-tag "$supervisorTag" \
-    ${BARYS_ARGUMENTS_VAR} \
-    --shared-downloads "$JENKINS_DL_DIR" \
-    --shared-sstate "$JENKINS_SSTATE_DIR" \
-    --rm-work
+docker stop yocto-build 2> /dev/null || true
+docker rm yocto-build 2> /dev/null || true
```
Page-: This should be `docker rm --volumes yocto-build 2> /dev/null || true` in order to make sure we clean up temporary volumes used by the container

automation/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+# Additional host packages required by resin
+RUN apt-get install -y jq npm nodejs sudo
+
+# Fix path of node/nodejs
+RUN ln -s /usr/bin/nodejs /usr/bin/node
+
+# Install docker
+# https://github.com/docker/docker/blob/master/project/PACKAGERS.md#runtime-dependencies
+RUN apt-get install -y iptables procps e2fsprogs xfsprogs xz-utils git kmod
+ENV DOCKER_VERSION 1.9.1
```
Page-: We need a `VOLUME /var/lib/docker` entry for docker in docker to play nicely

automation/jenkins_build.sh
```diff
@@ -26,25 +27,36 @@ fi
...
+docker run --rm -ti \
```
Page-: `-ti` is `tty` and `interactive` - the `interactive` at least probably doesn't make sense in an automated build?

automation/jenkins_build.sh
```diff
@@ -26,25 +27,36 @@ fi
...
+    -v $WORKSPACE:/yocto/resin-board \
+    -v $JENKINS_DL_DIR:/yocto/shared-downloads \
+    -v $JENKINS_SSTATE_DIR:/yocto/shared-sstate \
+    -e BUILDER_UID=$(id -u) \
+    -e BUILDER_GID=$(id -g) \
+    --name yocto-build \
+    --privileged\
```
Page-: Can you add a space before the \ please, just to match the other lines

automation/prepare-and-start.sh
```diff
@@ -0,0 +1,33 @@
...
+    else
+        echo "[ERROR] Timeout while waiting for docker to come up."
+        exit 1
+    fi
+done
+echo "[INFO] Docker was initialized."
+
+# Start barys with all the arguments requested
+echo "[INFO] Running build as builder user..."
+sudo -H -u builder /yocto/resin-board/resin-yocto-scripts/build/barys $@
```
Page-: After this finishes we should need to send a SIGTERM to docker and wait for it to exit before having the container exits so that docker can release the resources/loop devices it takes control of.

automation/Dockerfile
agherzan: I'll remove. Thanks.

automation/Dockerfile
agherzan: Sure. I'll improve as recommended.

automation/jenkins_build.sh
agherzan: Good catch. Will add.

automation/Dockerfile
agherzan: Fixing.

automation/jenkins_build.sh
agherzan: This was left from testing. I will remove.

automation/jenkins_build.sh
agherzan: Sure.

automation/prepare-and-start.sh
agherzan: Implementing and pushing.

Page-:
LGTM

automation/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+# Fix path of node/nodejs
+RUN ln -s /usr/bin/nodejs /usr/bin/node
+
+# Install docker
+# https://github.com/docker/docker/blob/master/project/PACKAGERS.md#runtime-dependencies
+RUN apt-get update && apt-get install -y iptables procps e2fsprogs xfsprogs xz-utils git kmod && rm -rf /var/lib/apt/lists/*
+ENV DOCKER_VERSION 1.10.1
+VOLUME /var/lib/docker
+RUN wget -q -O /usr/bin/docker https://get.docker.com/builds/Linux/x86_64/docker-${DOCKER_VERSION} && chmod +x /usr/bin/docker
```
petrosagg: Can we use a better method for downloading this executable? I'm not comfortable not using signed packages or at least binaries pinned with SHA hashes. @nghiant2710 can help you do this

agherzan: Added check for SHA hash.

telphan:
LGTM

automation/Dockerfile
Page-: I don't see the check?

agherzan: It was added as a later commit.

Page-: Oh you're right, I didn't realise I was in a specific commit (it looks identical to the usual diff - I can't spot anywhere other than the url that is different)

agherzan: I understood the confusion.

automation/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM ubuntu:15.04
+
+# Install the following utilities (required by poky)
+RUN apt-get update && apt-get install -y build-essential chrpath curl diffstat gcc-multilib gawk git-core libsdl1.2-dev texinfo unzip wget xterm cpio file && rm -rf /var/lib/apt/lists/*
+
```
floion: Can't the "rm -rf /var/lib/apt/lists/*" be after the last RUN command that uses apt-get? Do we have to remove it every time and then on the next RUN command to do "apt-get update" again and "rm -rf /var/lib/apt/lists/*" at the end of the RUN command?

automation/prepare-and-start.sh
```diff
@@ -0,0 +1,68 @@
...
+echo "[INFO] Creating and setting builder user $BUILDER_UID:$BUILDER_GID."
+groupadd docker
+groupadd -g $BUILDER_GID builder
+useradd -m -u $BUILDER_UID -g $BUILDER_GID -G docker builder
+
+# Start docker
+echo "[INFO] Starting docker."
+docker daemon 2> /dev/null &
+echo "[INFO] Waiting for docker to initialize..."
+STARTTIME=$(date +%s)
```
floion: guess STARTTIME should be evaluated before starting the docker daemon?

automation/prepare-and-start.sh
```diff
@@ -0,0 +1,68 @@
...
+cleanup() {
+    echo "[INFO] Running cleanup..."
+
+    # Stop docker gracefully
+    echo "[INFO] Stopping in container docker..."
+    DOCKERPIDFILE=/var/run/docker.pid
+    if [ -f $DOCKERPIDFILE ] && [ -s $DOCKERPIDFILE ] && ps $(cat $DOCKERPIDFILE) | grep -q docker; then
+        kill $(cat $DOCKERPIDFILE)
+        # Now wait for it to die
+        STARTTIME=$(date +%s)
```
floion: I guess STARTTIME should be evaluated before killing the docker daemon?

automation/prepare-and-start.sh
```diff
@@ -0,0 +1,68 @@
...
+# Start docker
+echo "[INFO] Starting docker."
+docker daemon 2> /dev/null &
+echo "[INFO] Waiting for docker to initialize..."
+STARTTIME=$(date +%s)
+ENDTIME=$(date +%s)
+until docker info >/dev/null 2>&1; do
+    if [ $(($ENDTIME - $STARTTIME)) -le $DOCKER_TIMEOUT ]; then
+        sleep 1
+        ENDTIME=$(date +%s)
```
floion: What if a SIGINT or SIGTERM arrives here? Then cleanup() would be invoked, and it would go through the

    else
 +        echo "[WARN] Can't stop docker container."
 +        echo "[WARN] Your host might have been left with unreleased resources (ex. loop devices)."
 +    fi

branch, without attempting to kill docker as it is not completely started yet. And it may start after this.
So what if you also pass cleanup() the amount of seconds left until DOCKER_TIMEOUT is reached so there is a chance for docker to start correctly and then to actually try to terminate it so that docker can free up his resources correctly?

automation/jenkins_build.sh
```diff
@@ -1,6 +1,22 @@
...
+    # Stop docker container
+    echo "[INFO] jenkins_build.sh: Cleaning up yocto-build container."
+    docker stop $BUILD_CONTAINER_NAME 2> /dev/null || true
+    docker rm --volumes $BUILD_CONTAINER_NAME 2> /dev/null || true
+
+    if [ "$1" == "fail" ]; then
+        exit 1
+    fi
+}
+trap 'cleanup fail' SIGINT SIGKILL
```
floion: Don't believe you can catch SIGKILL

automation/Dockerfile
Page-: It needs to be on every layer, because each `RUN` creates a docker layer and then the later layers are just overlaid on top, meaning that the moment a file is committed in one layer then you will always have to download it even if it's later deleted

agherzan: What Page said.

automation/jenkins_build.sh
agherzan: Good point. That was actually a typo initially. For whatever reason it remained here. I'll fix.

agherzan: It was supposed to be term.

automation/prepare-and-start.sh
agherzan: There is a catch here. The `docker info` command actually hangs making signal handlers blocked until docker either starts or fails. I'm not yet sure why this happens as I can't seem to reproduce on my computer. So when `docker wait` finishes, docker is either started or failed (completely). So the above case can't actually happen.

floion: Yes, if run synchronously then indeed, it cannot be interrupted. Which brings me to another thing: is there a chance "docker wait" may hang and sit there indefinitely? This may also cause this script to hang. You could run "docker wait" asynchronously and then wait for it to finish. This will ensure the script will not hang if docker wait hangs, but you also need to treat the above case I mentioned above.

agherzan: No it doesn't. It only hangs while it initializes.

floion: It doesn't as in "never hanged while you tested it"? Or as in "it cannot hang by design"?

agherzan: Never in my testing.

floion: It's still an external command that you call. That can nevertheless hang. Just a thought on this.

agherzan: The predecesor of the official docker in docker support (https://github.com/jpetazzo/dind/blob/master/wrapdocker#L103) does it too. So I don't think it will ever hang.

floion: If you are fine with it, then this PR is LGTM

PR #18, opened 2016-05-31
Various additions

agherzan:

CC @floion @Page- @petrosagg @telphan


floion:
"prepare-and-start.sh: Create first current group and then docker
If we do the other way around, docker group might take the GID of the one
the current goup has."

Typo above: goup instead of group

And also, you might want to rearrange "Create first current group" to "First create current group"

floion:
From the manpage of groupadd:

-g, --gid GID
           The numerical value of the group's ID. This value must be unique, unless the -o option is used. The value must be non-negative. The default is to use the smallest ID value greater than or equal
           to GID_MIN and greater than every other group.

           See also the -r option and the GID_MAX description.

So I am not sure exactly how the newly created group can be assigned the GID of the current group. Maybe the problem lies elsewhere?

agherzan:
Fixed the typos. As well I clarified the situation in the commit log. Tell me if you spot anything else.

telphan:
LGTM

petrosagg:
@agherzan why did we rename the dockerfile to a non-standard name?

automation/prepare-and-start.sh
```diff
@@ -37,8 +37,8 @@ trap 'cleanup fail' SIGINT SIGTERM
 
 # Create the normal user to be used for bitbake (barys)
 echo "[INFO] Creating and setting builder user $BUILDER_UID:$BUILDER_GID."
-groupadd docker
 groupadd -g $BUILDER_GID builder
```
floion: You could specify here "-f" along with "-g" and it will not depend on the order on how you create these groups. From the manpage of groupadd:

> -f, --force
> 
>            This option causes the command to simply exit with success status if the specified group already exists. When used with -g, and the specified GID already exists, another (unique) GID is chosen
>            (i.e.  -g is turned off).
> 

Then, when you user add, manpage of useradd says:

>        -g, --gid GROUP
>            The group name or number of the user's initial login group. The group name must exist. A group number must refer to an already existing group.
> 

So you can simply use the group name previously created.

floion:
LGTM

agherzan:
@petrosagg I did this for clarifying for what that Dockerfile is used. It should end up in the future in another repository but for now we can just have it here

agherzan:
As per no comments I'm going to merge this. If something will be raised, we can address them via additional PRs.

PR #21, opened 2016-06-26
jenkins_build.sh: --staging was replaced by --debug-image

agherzan:

@floion @telphan


floion:
LGTM

PR #22, opened 2016-07-14
jenkins_build.sh: No more debug images in non-production builds

agherzan:

### DIFF
--- a/automation/jenkins_build.sh
+++ b/automation/jenkins_build.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+
 set -e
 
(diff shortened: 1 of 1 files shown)


petrosagg:
Instead of just deleting this can we make it so that it gets its value from an env var that Jenkins provides? And then we'll make Jenkins default to non-debug. This way if someone wants a debug build it will be as simple as starting a jenkins job

agherzan:
Sounds like a good idea. I'll do that.

petrosagg:
SGTM

floion:
LGTM

Issue #23, opened 2016-07-18
Add ability to run tests if present in device repos


lifeeth:
Allow for having tests embedded in the device repositories with automation to run them during jenkins build.

Proposed directory structure is to have a directory called tests in the device repository root and have jenkins run tests/start.sh

PR #25, opened 2016-07-26
Add a env ENABLE_TESTS - that allows for disabling of tests or overriding in specific cases.

lifeeth:
Connects to #23


agherzan:
LGTM

PR #26, opened 2016-07-27
barys: Clarify invalid argument reason

agherzan:

### DIFF
--- a/build/barys
+++ b/build/barys
@@ -220,7 +220,7 @@ while [[ $# -ge 1 ]]; do
         --shared-downloads)
             # Argument needs to be non-empty and absolute path
             if [ -z "$2" ] || [ "${2:0:1}" != "/" ]; then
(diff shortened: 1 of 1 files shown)


build/barys
```diff
@@ -220,15 +220,15 @@ while [[ $# -ge 1 ]]; do
         --shared-downloads)
             # Argument needs to be non-empty and absolute path
             if [ -z "$2" ] || [ "${2:0:1}" != "/" ]; then
-                log ERROR "\"$1\" argument is invalid."
+                log ERROR "\"$1\" argument is invalid. Argument needs to be non-empty and absolute path."
```
Page-: I would reword this as `Argument needs to be a non-empty, absolute path` or as `Argument needs to be non-empty and an absolute path.`

agherzan: Done @Page- . Thanks.

PR #27, opened 2016-08-09
build-device-type-json.sh: make this work on ubuntu

michal-mazurek:
Some package managers install nodejs as 'node', while others as
'nodejs'.

Also change the wording in the error message a little.

Signed-off-by: Michal Mazurek <michal@resin.io>


floion:
LGTM

PR #28, opened 2016-08-10
build-device-type-json.sh: discard stderr when running "which"

michal-mazurek:
on some systems "which" prints to stderr if command is not found.

Signed-off-by: Michal Mazurek <michal@resin.io>


floion:
LGTM

PR #37, opened 2016-11-03
jenkins_build.sh: Don't allocate a pseudo-tty when running docker con…

agherzan:
…tainer

Not having a tty allocated and available, yocto build will revert to plain text
output.


petrosagg:
nice find!

agherzan:
Thanks @petrosagg . It was indeed a hard one!

PR #38, opened 2016-12-22
jenkins_build.sh: Upload production resinhup packages to registry.resinstaging.io

michal-mazurek:
connect https://github.com/resin-os/resinos/issues/155


michal-mazurek:
Any ideas how to test this without actually merging it? :)

PR #39, opened 2016-12-29
remove sstate duplicates after a build.

michal-mazurek:


michal-mazurek:
connect https://github.com/resin-os/resinos/issues/99

michal-mazurek:
We decided this might not work as expected, due to the fact that we build multiple different branches in jenkins. We are currently considering alternatives, like removing the entire dir once a month for example.

PR #41, opened 2017-01-19
Revert "jenkins_build.sh: Run resinio builds with this script"

floion:
This reverts commit de6ca4d930af5ad0a2aa11be6b0f5320369241dc.


floion:
We need this because in the 1.X branch there is no notion of --resinio or --resinos arguments for barys

PR #47, opened 2017-02-21
jenkins_build.sh: upload .bmap files to s3

michal-mazurek:
fix some whitespaces while here


michal-mazurek:
Will COMPRESSED or ARCHIVE invalidate the usefulness of a .bmap file? When are these options used? Can they be removed?

readlink --canonicalize seems to be GNU-only, but I think that's fine.

connect https://github.com/resin-os/resin-beaglebone/pull/71

michal-mazurek:
connect https://github.com/resin-os/resinos/issues/163

agherzan:
Needs rebasing but looks good.

lurch:
ping @jhermsmeier since I guess this would be useful for Etcher too?

telphan (approved):

PR #56, opened 2017-04-14
IGNORE file in build folder while files are being uploaded

zvin:
Upload an empty IGNORE file before uploading image files, remove it once the upload is finished.
This will allow the image maker to ignore incomplete versions.

[related image maker PR](https://github.com/resin-io/resin-image-maker/pull/73)

### DIFF
--- a/automation/jenkins_build.sh
+++ b/automation/jenkins_build.sh
@@ -218,14 +218,23 @@ if [ "$deploy" == "yes" ]; then
 		-e SLUG="$SLUG" \
 		-e BUILD_VERSION="$S3_VERSION_HOSTOS" \
 		-e DEVELOPMENT_IMAGE="$DEVELOPMENT_IMAGE" \
-		-v $S3_DEPLOY_DIR:/host/images resin/resin-img:master /bin/sh -x -c ' \
-		echo "${BUILD_VERSION}" > "/host/images/${SLUG}/latest" \
-		&& /usr/src/app/node_modules/.bin/coffee /usr/src/app/scripts/prepare.coffee \
-		&& apt-get -y update \
-		&& apt-get install -y s3cmd \
-		&& ([ "${DEVELOPMENT_IMAGE}" = "no" ] && $S3_CMD put /host/images/${SLUG}/latest s3://${S3_BUCKET}/${SLUG}/ || true) \
-		&& ([ -z "$($S3_CMD ls s3://${S3_BUCKET}/${SLUG}/${BUILD_VERSION}/)" ] \
-		&& $S3_CMD $S3_SYNC_OPTS sync /host/images/${SLUG}/${BUILD_VERSION}/ s3://${S3_BUCKET}/${SLUG}/${BUILD_VERSION}/)'
+		-v $S3_DEPLOY_DIR:/host/images resin/resin-img:master /bin/sh -x -e -c ' \
+			echo "${BUILD_VERSION}" > "/host/images/${SLUG}/latest"
+			/usr/src/app/node_modules/.bin/coffee /usr/src/app/scripts/prepare.coffee
+			apt-get -y update
+			apt-get install -y s3cmd
+			if [ -z "$($S3_CMD ls s3://${S3_BUCKET}/${SLUG}/${BUILD_VERSION}/)" ]; then
+				touch IGNORE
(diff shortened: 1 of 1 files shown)


automation/jenkins_build.sh
```diff
@@ -218,14 +218,21 @@ if [ "$deploy" == "yes" ]; then
...
+			/usr/src/app/node_modules/.bin/coffee /usr/src/app/scripts/prepare.coffee
+			apt-get -y update
+			apt-get install -y s3cmd
+			if [ -z "$($S3_CMD ls s3://${S3_BUCKET}/${SLUG}/${BUILD_VERSION}/)" ]; then
+				touch IGNORE
+				$S3_CMD put IGNORE s3://${S3_BUCKET}/${SLUG}/${BUILD_VERSION}/
+				$S3_CMD $S3_SYNC_OPTS sync /host/images/${SLUG}/${BUILD_VERSION}/ s3://${S3_BUCKET}/${SLUG}/${BUILD_VERSION}/
+				$S3_CMD rm s3://${S3_BUCKET}/${SLUG}/${BUILD_VERSION}/IGNORE
+			fi
+			if [ "${DEVELOPMENT_IMAGE}" = "no" ]; then
```
agherzan: I think this is should be inside the above one as we don't want the latest to be set if we haven't upload the version. As well, if the version is already there the build should fail.

zvin: updated

automation/jenkins_build.sh
```diff
@@ -218,14 +218,21 @@ if [ "$deploy" == "yes" ]; then
...
+		-v $S3_DEPLOY_DIR:/host/images resin/resin-img:master /bin/sh -x -e -c ' \
+			echo "${BUILD_VERSION}" > "/host/images/${SLUG}/latest"
+			/usr/src/app/node_modules/.bin/coffee /usr/src/app/scripts/prepare.coffee
+			apt-get -y update
+			apt-get install -y s3cmd
+			if [ -z "$($S3_CMD ls s3://${S3_BUCKET}/${SLUG}/${BUILD_VERSION}/)" ]; then
+				touch IGNORE
+				$S3_CMD put IGNORE s3://${S3_BUCKET}/${SLUG}/${BUILD_VERSION}/
+				$S3_CMD $S3_SYNC_OPTS sync /host/images/${SLUG}/${BUILD_VERSION}/ s3://${S3_BUCKET}/${SLUG}/${BUILD_VERSION}/
+				$S3_CMD rm s3://${S3_BUCKET}/${SLUG}/${BUILD_VERSION}/IGNORE
```
agherzan: Shouldn't this come at the end? After the latest put.

zvin: I'm not sure.
IMO `latest` should point to an image that is ready: without the `IGNORE` file.

agherzan: I think you have a point there.

automation/jenkins_build.sh
```diff
@@ -218,14 +218,21 @@ if [ "$deploy" == "yes" ]; then
...
-		&& apt-get install -y s3cmd \
-		&& ([ "${DEVELOPMENT_IMAGE}" = "no" ] && $S3_CMD put /host/images/${SLUG}/latest s3://${S3_BUCKET}/${SLUG}/ || true) \
-		&& ([ -z "$($S3_CMD ls s3://${S3_BUCKET}/${SLUG}/${BUILD_VERSION}/)" ] \
-		&& $S3_CMD $S3_SYNC_OPTS sync /host/images/${SLUG}/${BUILD_VERSION}/ s3://${S3_BUCKET}/${SLUG}/${BUILD_VERSION}/)'
+		-v $S3_DEPLOY_DIR:/host/images resin/resin-img:master /bin/sh -x -e -c ' \
+			echo "${BUILD_VERSION}" > "/host/images/${SLUG}/latest"
+			/usr/src/app/node_modules/.bin/coffee /usr/src/app/scripts/prepare.coffee
+			apt-get -y update
+			apt-get install -y s3cmd
+			if [ -z "$($S3_CMD ls s3://${S3_BUCKET}/${SLUG}/${BUILD_VERSION}/)" ]; then
```
agherzan: What happens if this if doesn't match? Will the build fail? In other words, will the build fail if the version it tries to deploy is already in s3? It was doing that before.

zvin: It would just do nothing.
I added an else with `exit 1`

PR #59, opened 2017-04-26
jenkins_build.sh: resin-img generates files in the name of the curren…

agherzan:
…t user

In this way we avoid any permissions errors afterwards.

### DIFF
--- a/automation/jenkins_build.sh
+++ b/automation/jenkins_build.sh
@@ -222,22 +222,29 @@ deploy_to_s3() {
 		-e SLUG="$SLUG" \
 		-e BUILD_VERSION="$_s3_version_hostos" \
 		-e DEVELOPMENT_IMAGE="$DEVELOPMENT_IMAGE" \
+		-e DEPLOYER_UID=$(id -u) \
+		-e DEPLOYER_GID=$(id -g) \
 		-v $_s3_deploy_dir:/host/images resin/resin-img:master /bin/sh -x -e -c ' \
-			echo "${BUILD_VERSION}" > "/host/images/${SLUG}/latest"
-			/usr/src/app/node_modules/.bin/coffee /usr/src/app/scripts/prepare.coffee
 			apt-get -y update
 			apt-get install -y s3cmd
-			if [ -z "$($S3_CMD ls s3://${S3_BUCKET}/${SLUG}/${BUILD_VERSION}/)" ]; then
-				touch IGNORE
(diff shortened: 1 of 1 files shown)


automation/jenkins_build.sh
```diff
@@ -218,11 +218,17 @@ if [ "$deploy" == "yes" ]; then
...
 		-v $S3_DEPLOY_DIR:/host/images resin/resin-img:master /bin/sh -x -e -c ' \
+			echo "Creating and setting deployer user $DEPLOYER_UID:$DEPLOYER_GID."
+			groupadd -g $DEPLOYER_GID deployer
+			useradd -m -u $DEPLOYER_UID -g $DEPLOYER_GID deployer
+			sudo su deployer
 			echo "${BUILD_VERSION}" > "/host/images/${SLUG}/latest"
 			/usr/src/app/node_modules/.bin/coffee /usr/src/app/scripts/prepare.coffee
-			apt-get -y update
-			apt-get install -y s3cmd
+			sudo apt-get -y update
```
zvin: You can move this above  `sudo su deployer` and avoid sudo

agherzan: Correct. I'm wondering if there is any gain from it - as it's still two lines change.

automation/jenkins_build.sh
```diff
@@ -218,22 +218,29 @@ if [ "$deploy" == "yes" ]; then
 		-e SLUG="$SLUG" \
 		-e BUILD_VERSION="$S3_VERSION_HOSTOS" \
 		-e DEVELOPMENT_IMAGE="$DEVELOPMENT_IMAGE" \
+		-e DEPLOYER_UID=$(id -u) \
```
Page-: I think you can just do `--user="$(id -u):$(id -g)` rather than having to manually create/su to the user: https://docs.docker.com/engine/reference/run/#user

agherzan: `The image developer can create additional users. Those users are accessible by name.` This sounds like you actually need to create them first in the image. But then you need to install packages which you need to do with sudo so if we gain the ability of not needing to su we will need to include sudo and use it for installing the other packages. Actually this won't work at all. Because sudo is not installed in the image and installing it needs root.

Page-: You stopped reading one sentence too soon @agherzan, the very next sentence is:
> When passing a numeric ID, the user does not have to exist in the container.

:P

PR #60, opened 2017-05-04
Allow cloning of private git repos from within the container

floion:


automation/jenkins_build.sh
```diff
@@ -119,6 +119,8 @@ docker run --rm \
     -v $WORKSPACE:/yocto/resin-board \
     -v $JENKINS_DL_DIR:/yocto/shared-downloads \
     -v $JENKINS_SSTATE_DIR:/yocto/shared-sstate \
+    -v $SSH_AUTH_SOCK:/tmp/ssh-agent \
+    -e SSH_AUTH_SOCK=/tmp/ssh-agent \
```
willnewton: This seems reasonable although I do worry about what keys are available via this socket. I guess if we are building partially untrusted code (some random guy from the internet's fork of our code) we would be exposing keys to that build process.

PR #73, opened 2017-06-07
Attempt to fix unrecognized input header error

brownjohnf:
Based on the threads listed below, this should resolve the issue, which
is related to building an image with the interactive terminal flag (-t)
but then running it without the flag.

https://github.com/moby/moby/pull/4882
https://github.com/moby/moby/issues/4857
https://github.com/moby/moby/issues/5327

Change-Type: patch

### DIFF
--- a/automation/jenkins_build.sh
+++ b/automation/jenkins_build.sh
@@ -218,7 +218,7 @@ deploy_to_s3() {
 
 	local _s3_cmd="s3cmd --access_key=${_s3_access_key} --secret_key=${_s3_secret_key}"
 	local _s3_sync_opts="--recursive --acl-public"
-	docker run \
+	docker run -it \
 		-e BASE_DIR=/host/images \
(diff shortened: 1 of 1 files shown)


floion:
lgtm
@agherzan to clarify on why the intercative mode was specified in the first place for this, if for any reason

agherzan:
Where is this interactive mode specified?

brownjohnf:
I'm assuming that somewhere in the inheritance chain an image was built with interactive mode. Read those threads for more info, but this is just my best guess, at this point.

brownjohnf:
Er, rather, allocating a pseudo-tty. I don't think the -i flag is strictly necessary, but i usually use them together. It's also possible that s3cmd just misbehaves when it's run without a tty or something.

agherzan:
As far as I know we don't build images with interactive mode. I don't even know how that is possible?

PR #75, opened 2017-06-14
jenkins_build.sh: Remove pseudo tty allocation

floion:
Having this arg for docker run produces the following error:
cannot enable tty mode on non tty input


willnewton:
This looks like it may have some left over debug code in it?

floion:
Hey Will, that was some testing I was doing on the PR.
Only now it can be reviwed.

floion:
Replaced by #76

Issue #74, opened 2017-06-09
resin/resin-img doesn't clean up the container after running it


agherzan:

floion:
fixed by #75

floion:
Fix merged. Closing this issue

PR #76, opened 2017-06-15
jenkins_build.sh: Fix issues with "docker run" of resin/resin-img

floion:
We remove the "-i" arg as that throws the following error:

"cannot enable tty mode on non tty input"

We also add the "--rm" flag to automatically remove the container when it exits
as it makes no sense for it to be running after exiting.


floion:
fixes #74

PR #82, opened 2017-06-30
Dockerfile_yocto-build-env: Do not include SDL

floion:
Yocto Morty can build a libsdl-native so there is no need for it in the docker image.

Also we install python 2.7 as it is explicitly needed by some packages. E.g. xcb-proto-native


willnewton:
Is this ok for the boards not on morty yet?

floion:
Yeah it should be, this is only needed for Qemu builds and our qemu builds are on Morty

PR #85, opened 2017-07-26
pass the json file as argument to jq

michal-mazurek:


build/barys
```diff
@@ -444,11 +444,11 @@ else
   fi
   for machine in $ITERATOR_MACHINES; do
       for json in ${DEVICE_TYPES_JSONS}; do
-        MACHINE_IN_JSON=`cat $json | jq  -r '.yocto.machine'`
+        MACHINE_IN_JSON=`jq -r '.yocto.machine' $json`
         if [[ "$machine" != "${MACHINE_IN_JSON}" ]]; then
```
willnewton: Also no need for double brackets here.

PR #88, opened 2017-08-21
build/barys: Set supervisor tag last

willnewton:
Set the supervisor tag value in local.conf after settig the
default value in order to avoid the value being overwritten.

### DIFF
--- a/build/barys
+++ b/build/barys
@@ -396,9 +396,6 @@ else
     # Always activate resinhup bundles in production builds
     sed -i "s/.*RESINHUP ?=.*/RESINHUP ?= \"yes\"/g" conf/local.conf
 fi
-if [ -n "$SUPERVISOR_TAG" ]; then
-    sed -i "s/.*TARGET_TAG ?=.*/TARGET_TAG ?= \"$SUPERVISOR_TAG\"/g" conf/local.conf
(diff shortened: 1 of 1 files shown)


build/barys
```diff
@@ -396,9 +396,6 @@ else
     # Always activate resinhup bundles in production builds
     sed -i "s/.*RESINHUP ?=.*/RESINHUP ?= \"yes\"/g" conf/local.conf
 fi
-if [ -n "$SUPERVISOR_TAG" ]; then
-    sed -i "s/.*TARGET_TAG ?=.*/TARGET_TAG ?= \"$SUPERVISOR_TAG\"/g" conf/local.conf
-fi
```
floion: Who can overwrite this?

willnewton: Further down in this script the value gets overwritten.

agherzan: I can't find where this variable was overwritten either. Could you point us to a code line?

willnewton: https://github.com/resin-os/resin-yocto-scripts/blob/master/build/barys#L419
https://github.com/resin-os/resin-yocto-scripts/blob/master/build/barys#L425

The variable in question is TARGET_TAG

PR #89, opened 2017-08-24
build/barys: Add --build-history flag

willnewton:
Allow enabling the Yocto build history feature from barys.


willnewton:
@agherzan it was inspired by the need to do size profiling for https://github.com/resin-os/meta-resin/issues/810

PR #90, opened 2017-08-31
jenkins_build.sh: Allow non-jenkins docker container builds

floion:
jenkins_build.sh should also be able to run outside of the jenkins environment
and run the yocto build in a docker container.


willnewton:
This looks ok but I think we will also need a version of this on the 1.X branch too as it changes the calling convention of the script. Would it be possible to preserve backwards compatibility? It might make things easier in that regard.

floion:
Do we want to have this ability on the 1.X line?

PR #93, opened 2017-09-06
automation/jenkins_yocto-build-env.sh: Get revision directly

willnewton:
Get the current short commit ID by calling git rather than
relying on Jenkins to inject the variable.

### DIFF
--- a/automation/jenkins_yocto-build-env.sh
+++ b/automation/jenkins_yocto-build-env.sh
@@ -3,16 +3,12 @@
 set -ev
 
 DOCKERFILE=Dockerfile_yocto-build-env
+REVISION=$(git rev-parse --short HEAD)
 
 # Get the absolute script location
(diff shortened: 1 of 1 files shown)


automation/jenkins_yocto-build-env.sh
```diff
@@ -3,17 +3,13 @@
 set -ev
 
 DOCKERFILE=Dockerfile_yocto-build-env
+REVISION=$(git rev-parse --short HEAD)
```
brownjohnf: Do you need to export this variable as well? I guess I'm sure where all you use this value. This will also depend on the git context you're in, and I'm not sure how that's structured either. If this is just for building this repo, then it should be fine.

agherzan: I think we are in the workdir when running this script while the repository is cloned in a subdirectory.

PR #110, opened 2018-02-01
Allow to build a custom "production" resinos image

bbinet:
We can already set the `--development-image` flag for barys to build a development image

### DIFF
--- a/build/barys
+++ b/build/barys
@@ -434,7 +434,6 @@ else
     sed -i "s/.*RESIN_CONNECTABLE_ENABLE_SERVICES ?=.*/RESIN_CONNECTABLE_ENABLE_SERVICES ?= \"0\"/g" conf/local.conf
     sed -i "s/.*TARGET_REPOSITORY ?=.*/TARGET_REPOSITORY ?= \"\"/g" conf/local.conf
     sed -i "s/.*TARGET_TAG ?=.*/TARGET_TAG ?= \"\"/g" conf/local.conf
-    sed -i "s/.*DEVELOPMENT_IMAGE =.*/DEVELOPMENT_IMAGE = \"1\"/g" conf/local.conf
 fi
 if [ -n "$SUPERVISOR_TAG" ]; then
     sed -i "s/.*SUPERVISOR_TAG ?=.*/SUPERVISOR_TAG ?= \"$SUPERVISOR_TAG\"/g" conf/local.conf


build/barys
```diff
@@ -434,7 +434,6 @@ else
     sed -i "s/.*RESIN_CONNECTABLE_ENABLE_SERVICES ?=.*/RESIN_CONNECTABLE_ENABLE_SERVICES ?= \"0\"/g" conf/local.conf
     sed -i "s/.*TARGET_REPOSITORY ?=.*/TARGET_REPOSITORY ?= \"\"/g" conf/local.conf
     sed -i "s/.*TARGET_TAG ?=.*/TARGET_TAG ?= \"\"/g" conf/local.conf
-    sed -i "s/.*DEVELOPMENT_IMAGE =.*/DEVELOPMENT_IMAGE = \"1\"/g" conf/local.conf
```
floion: Having this in makes it that a resinos build by default is a development image. So that is intended. Is this behavior not working for you for some reason?

bbinet: Yes, how can I build a (non resinio) image which is still a production one, so that the serial console is disabled?

floion: You can use https://github.com/resin-os/resin-yocto-scripts/blob/master/automation/jenkins_build.sh

This allows you to build one of the 4 build types: managed-dev, managed-prod, unmanaged-dev, unmanaged-prod. So you would run this script with the "-b unmanaged-prod" switch

bbinet: Can you check the `unmanaged-prod` is actually a prod image?
When I look at the code, `jenkins_build.sh` calls `prepare-and-start.sh` script through docker which finally calls `barys` which will force the `DEVELOPMENT_IMAGE` variable to be set to 1 for all unmanaged builds.

floion: Indeed @bbinet , I am checking to see if there is a particular reason why we make all non-resin builds default to development flags.
@agherzan and @shaunmulligan please chime in if you remember the reasoning for this

shaunmulligan: @bbinet and @floion the default to build "development" images is largely due to the fact that today the pre-built resinOS unmanaged images are not very useful when in "production" mode as there is no easy way for a user to get a container on to the device, since there is not way to start preloaded containers (since now resin supervisor) and there is no SSH and docker socket connection. So there is no really good way to use a pre-built resinOS production image. 

Fortunately there is some new work by @agherzan which is to address this and make resinOS prod images more usable.

bbinet: @shaunmulligan @floion https://github.com/resin-os/meta-resin/#resin_connectable--0 says that we can use the TARGET_IMAGE and TARGET_TAG variable to load a custom image.
Personnaly I'm building my own resinos image, and I had to use this patch to get a production image.

bbinet: @shaunmulligan @floion since barys already provides a `--development-image` option that can be used to trigger a development build whenever we need, I think it does not make much sense to force a development image on non-resinio builds.

floion: I am okay with this PR. @petrosagg how do you feel about this?

PR #112, opened 2018-02-27
Allow additionnal variables that contains "-" char

bbinet:
This is useful for example to add variables suffixed with machine name, for example:
TARGET_REPOSITORY_genericx86-64=my.docker.repo/custom/targetimage


floion:
Thanks for the contribution.
Merged

PR #113, opened 2018-03-22
jenkins_build.sh: Add support for device types that don't deploy raw images

agherzan:

### DIFF
--- a/automation/jenkins_build.sh
+++ b/automation/jenkins_build.sh
@@ -60,6 +60,11 @@ deploy_build () {
 
 	test "$SLUG" = "edge" && return
 
+	if [ -z "$_deploy_artifact" ] && [ -z "$_deploy_flasher_artifact" ]; then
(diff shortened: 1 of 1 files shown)


automation/jenkins_build.sh
```diff
@@ -60,6 +60,11 @@ deploy_build () {
 
 	test "$SLUG" = "edge" && return
 
+	if [ -z "$_deploy_artifact" ] && [ -z "$_deploy_flasher_artifact" ]; then
+		echo "[WARN] No deploy artifacts defined."
+		return
+	fi
+
 	cp -v "$YOCTO_BUILD_DEPLOY/kernel_modules_headers.tar.gz" "$_deploy_dir"
```
floion: Do these devices don't create kernel modules headers archive?

agherzan: These device types are meant to run only in a container so they rely on the host's kernel.

PR #122, opened 2018-07-21
jenkins_build.sh: Deploy to both resinio and resinos S3 buckets

floion:
We now have unified builds for managed and unmanaged. So the decision was that for
each build, we deploy to both S3 buckets for resinio and resinos at the same time.

### DIFF
--- a/automation/jenkins_build.sh
+++ b/automation/jenkins_build.sh
@@ -286,6 +286,7 @@ deploy_resinhup_to_registries() {
 }
 
 deploy_to_s3() {
+	local _s3_bucket=$1
 	local _s3_version_hostos=$VERSION_HOSTOS
 	if [ "$DEVELOPMENT_IMAGE" = "yes" ]; then
 		_s3_version_hostos=$_s3_version_hostos.dev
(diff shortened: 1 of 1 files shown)


automation/jenkins_build.sh
```diff
@@ -297,16 +297,18 @@ deploy_to_s3() {
 
 	deploy_build "$_s3_deploy_images_dir" "false"
```
agherzan: I think it would easier and cleaner to make the bucket a function parameter and just call the function (deploy_to_s3) twice.

agherzan: That's what I propose. As in this way we can easily remove this in the future - we will stop deploying in two places hopefully soon.

automation/jenkins_build.sh
```diff
@@ -364,7 +363,16 @@ if [ "$deploy" = "yes" ]; then
...
-	deploy_to_s3
+	if [ "$deployTo" = "production" ]; then
+		S3_BUCKET_RESINIO=resin-production-img-cloudformation/images
+		S3_BUCKET_RESINOS=resin-production-img-cloudformation/resinos
+	elif [ "$deployTo" = "staging" ]; then
+		S3_BUCKET_RESINIO=resin-staging-img/images
+		S3_BUCKET_RESINOS=resin-staging-img/resinos
+	fi
+
+	deploy_to_s3 $S3_BUCKET_RESINIO
```
agherzan: Double quote to prevent globbing and word splitting.

automation/jenkins_build.sh
```diff
@@ -364,7 +363,16 @@ if [ "$deploy" = "yes" ]; then
...
+	deploy_to_s3 $S3_BUCKET_RESINOS
```
agherzan: Double quote to prevent globbing and word splitting.

PR #124, opened 2018-08-07
Dockerfile_yocto-build-env: Add libssl-dev

floion:
This package is needed for compiling kernel-modules-headers
on newer kernels (encountered on 4.16)

Just adding openssl-native as a dependency of kernel-modules-headers
does not fix the issue. The headers get installed in the sysroot of
kernel-modules-headers but compilation of the scripts fails
nonetheless.

Solution is to add libssl-dev in the container so that at build time
openssl/bio.h is found in the host.


floion:
retest

floion:
Fixed by https://github.com/resin-os/module-headers/pull/18

PR #130, opened 2018-12-14
Switch from resin-yocto-scripts to balena-yocto-scripts

floion:


floion:
@resin-ci retest

PR #136, opened 2019-04-09
jenkins_build.sh: Replace meta-resin reference with meta-balena

floion:
The meta-resin layer is now renamed to meta-balena

Change-type: patch


floion:
@resin-ci retest

PR #153, opened 2019-09-10
jenkins_build.sh: Add kernel_modules_headers_src.tar.gz

ZubairLK:
Merge after https://github.com/balena-os/meta-balena/pull/1605

That PR adds another artifact which is the kernel headers source tarball.


ZubairLK:
This needs more thought and a discussion in an arch call

PR #155, opened 2019-09-17
Test balena deploy

nazrhom:


automation/jenkins_build.sh
```diff
@@ -365,6 +377,9 @@ EOSU
 }
 
 # Deploy
+# Test deploy
+deploy_resinhup_to_registries
```
floion: Why is this done here without any check? This same function is also called bellow in a check if the deploy flag is set

automation/jenkins_build.sh
```diff
@@ -289,10 +289,22 @@ deploy_resinhup_to_registries() {
 
 	local _hostapp_image=$(docker load --quiet -i "$_resinhup_path" | cut -d: -f1 --complement | tr -d ' ')
 	docker tag "$_hostapp_image" "$_docker_repo:$_tag"
-	docker push $_docker_repo:$_tag
+	# docker push $_docker_repo:$_tag
```
floion: Can this be removed completely if it is intended not to be used anymore?

automation/start-docker-and-push.sh
```diff
@@ -0,0 +1,80 @@
...
+    fi
+done
+echo "[INFO] Docker was initialized."
+
+_local_image=$(docker load -i /host/resin-image.docker | cut -d: -f1 --complement | tr -d " " )
+
+echo "[INFO] Logging into $deployTo as balenaos"
+if [ "$DEPLOY_TO" = "staging" ]; then
+	export BALENARC_BALENA_URL=balena-staging.com
+  echo $BALENAOS_STAGING_TOKEN
```
floion: This probably is a testing artifact?

PR #157, opened 2019-10-08
Test balena deploy

nazrhom:


automation/Dockerfile_yocto-build-env
```diff
@@ -30,13 +30,6 @@ VOLUME /var/lib/docker
 RUN wget -q -O /usr/bin/docker https://get.docker.com/builds/Linux/x86_64/docker-${DOCKER_VERSION} && chmod +x /usr/bin/docker \
     && echo "${DOCKER_SHA256}" /usr/bin/docker | sha256sum -c -
 
-# Install balena-cli
```
floion: Why was the cli removed?

nazrhom: Because it was never needed in that image, the balena-push-env container now contains the CLI and is used to push the release

PR #168, opened 2020-02-28
jenkins_build.sh: Allow for passing additional arguments to barys

floion:
This is useful for passing SD8887_FROM_SOURCE for example so that
the Fin build will compile the wifi and bt driver from source.

Change-type: patch


floion:
@balena-ci retest

PR #171, opened 2020-05-08
Manage deploying private device types

sradevski:
Set S3 access policy to private for private DT
and don't deploy private DT to dockerhub

This is untested (and I don't know how this script gets tested),
but I believe these are the only changes necessary to support private DTs.
Let's sync in chat on testing this, I just wanted to get the ball rolling.

Change-type: minor
Signed-off-by: Stevche Radevski <stevche@balena.io>


floion:
@balena-ci retest

PR #176, opened 2020-10-21
build-device-type-json: show stderr on failure

jakogut:
When building a device type JSON manifest, any syntax error is handled
by printing a message about node being unavailable, even when it's
installed and working.

Check that node is installed and executable before building the
manifest, and exit early with the existing message if not. Otherwise,
print stderr to the console for debugging.


alexgg (changes requested):
@jakogut it's missing the footers (Change-type, Signed-off-by: etc) take a look at the commit guidelines and past commits to get an idea https://github.com/balena-io/process/blob/master/process/Commit_and_PR_Guidelines.md#commit-guidelines

jakogut:
@alexgg Thanks for the feedback, fixed.

floion (approved):

alexgg (approved):

PR #181, opened 2021-02-04
jenkins: deploy test suite from meta-balena

klutchell:
If leviathan test suites are found in the checkout of
meta-balena we should package them up with the jenkins artifacts.

Change-type: patch
Connects-to: https://github.com/balena-os/meta-balena/pull/2113

### DIFF
--- a/automation/jenkins_build.sh
+++ b/automation/jenkins_build.sh
@@ -111,6 +111,12 @@ deploy_build () {
 			rm -rf $_deploy_dir/image/resin-flasher.img
 		fi
 	fi
+
+	if [ -d "${WORKSPACE}/layers/meta-balena/tests" ]
+	then
+		# copy all leviathan/testbot tests from meta-balena to the deploy dir
+		cp -av "${WORKSPACE}/layers/meta-balena/tests" "$_deploy_dir/tests"
+	fi
 }
 
 rootdir="$( cd "$( dirname "$0" )" && pwd )/../../"


vipulgupta2048:
@klutchell I would still like to test  
1. Why we are cloning meta-balena? Are we actually cloning it?
2. If we are cloning meta-balena, then the tests are already there in the build. We just need to redirect them. 
3. If not then this looks good, I will make that PR to meta-balena and verify this on my local pipeline first

klutchell:
@vipulgupta2048 

>     1. Why we are cloning meta-balena? Are we actually cloning it?
> 
>     2. If we are cloning meta-balena, then the tests are already there in the build. We just need to redirect the primary leviathan config to that folder. (Bad solution as this might result in new tests not being taken into account)
> 
>     3. If not then this looks good, I will make that PR to meta-balena and verify this on my local pipeline first

It looks like we clone meta-balena to get the commit hash for the downstream [leviathan-set-meta-balena-status](https://jenkins.dev.resin.io/job/leviathan-set-meta-balena-status) job.

```bash
# Resolve meta-balena commit if the branch is spcified.
if [ -n "$metaResinBranch" ] && [ "$metaResinBranch" != "__ignore__" ]; then
  git clone https://github.com/balena-os/meta-balena.git .meta-balena
  metaBalenaCommit=$(cd .meta-balena && git rev-parse origin/$metaResinBranch)
  echo "Resolving meta-balena revision for $metaResinBranch => $metaBalenaCommit"
  echo "ghprbActualCommit=$metaBalenaCommit" > .meta-balena-commit
fi
```

It's a full clone for now so yes, we could just point to `./meta-balena/tests` in the Leviathan workspace to get the suite of tests.

However, I think reading them from the artifact so we can treat the tests as an "output" of a build is a better way to proceed. The checkout of meta-balena may not always be present based on the conditions above, and we may change it to a shallow clone at some point since it's only purpose is to get a commit hash.

vipulgupta2048 (commented):
+1 to shallow cloning, keeping tests as artefacts and getting them as output.
LGTM from my side

klutchell:
@vipulgupta2048 Let's wait to merge this until the test suite has been added to meta-balena. I wrote it to check that `meta-balena/tests` exist before trying to add to artifacts, but to avoid extra churn if that path changes or or additional packaging is required we should keep this in draft I think.

alexgg (approved):
lgtm

PR #182, opened 2021-02-04
esr: Add branch creation automation script

alexgg:
Change-type: patch

### DIFF
--- a/esr/esr_release.py
+++ b/esr/esr_release.py
@@ -1,0 +1,493 @@
+#!/usr/bin/env python3
+
+import sys, os, re, contextlib, yaml, tempfile, git, logging, time, urllib.request, json, requests
+from getopt import getopt
+from getopt import GetoptError
+from subprocess import Popen,PIPE
+from distutils.version import LooseVersion
+from datetime import datetime
+
+logger = logging.getLogger()
+
+class DeviceRepo:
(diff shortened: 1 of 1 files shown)


esr/esr_release.py
```diff
@@ -0,0 +1,493 @@
...
+    lf.setLevel(logging.DEBUG)
+    lf.setFormatter(formatter)
+    logger.addHandler(lf)
+    ch = logging.StreamHandler()
+    ch.setLevel(logging.INFO)
+    ch.setFormatter(formatter)
+    logger.addHandler(ch)
+
+    try:
+        opts, args = getopt(argv[1:], "he:b:u:t:d:a:g:k", ["help", "esr-version=", "bos-version=", "user", "token", "deploy-to", "api-token", "github-api-token", "keep-tmp-dir="])
```
floion, line 268: Shouldn't the long style options that require an argument be suffixed with `=`?

alexgg: just the ones that require an argument, but I missed some :) thanks

alexgg:
For the record, we considered moving this script somewhere else as it won't be used by the device build system but the CI, but at the same time there is no better place for OS related CI scripts, and with the new architecture this repository should hopefully not be needed at device built time and could move to storing CI automation.

alexgg:
@balena-ci I self-certify!

PR #185, opened 2021-02-23
jenkins: compress test suites inside jenkins artifacts

klutchell:
The jenkins leviathan builds will attempt to flatten these artifacts on extract (`unzip -j`) so make sure the test suites are a separate tarball inside the artifacts.

Change-type: patch

### DIFF
--- a/automation/jenkins_build.sh
+++ b/automation/jenkins_build.sh
@@ -114,8 +114,9 @@ deploy_build () {
 
 	if [ -d "${WORKSPACE}/layers/meta-balena/tests" ]
 	then
-		# copy all leviathan/testbot tests from meta-balena to the deploy dir
-		cp -av "${WORKSPACE}/layers/meta-balena/tests" "$_deploy_dir/tests"
+		# package all leviathan/testbot tests from meta-balena to the deploy dir
+		# make sure they are compressed so a flattened unzip of artifacts does not fail
+		tar -czvf "$_deploy_dir/tests.tar.gz" "${WORKSPACE}/layers/meta-balena/tests"
 	fi
 }
 


automation/jenkins_build.sh
```diff
@@ -115,7 +115,7 @@ deploy_build () {
 	if [ -d "${WORKSPACE}/layers/meta-balena/tests" ]
 	then
 		# copy all leviathan/testbot tests from meta-balena to the deploy dir
-		cp -av "${WORKSPACE}/layers/meta-balena/tests" "$_deploy_dir/tests"
+		tar -czvf "$_deploy_dir/tests.tar.gz" "${WORKSPACE}/layers/meta-balena/tests"
```
vipulgupta2048, line 118: - We need to change the comment and 
- We need to copy the tests.tar.gz to the `$_deploy_dir` as well?

vipulgupta2048 (commented):
LGTM

automation/jenkins_build.sh
klutchell: The file path for the output tarball is already in the deploy dir (`"$_deploy_dir/tests.tar.gz"`)

klutchell:
Also note that this change is being made to the jenkins leviathan template to handle all cases in the interim.

**current:**
```bash
unzip -j artifacts.zip -d .leviathan-inputs
( cd .leviathan-inputs && unzip balena.img.zip && gzip balena.img )
```
**proposed:**
```bash
# handle temporary cases where tests were not compressed and could not be flattened
if ! unzip -j artifacts.zip -d .leviathan-inputs
then
	# unzip normally with subfolders and move everything up one level
	unzip artifacts.zip -d .leviathan-inputs
    ( cd .leviathan-inputs && mv -v deploy-jenkins/* ./)
fi
    
( cd .leviathan-inputs && unzip balena.img.zip && gzip balena.img )

# unpack test suites if they exist
if [ -f .leviathan-inputs/tests.tar.gz ]
then
    ( cd .leviathan-inputs && tar -xzvf tests.tar.gz )
fi
```

alexgg (approved):

PR #186, opened 2021-02-25
Parametrize repository names and other generic changes

alexgg:
* Parametrize application, target image and docker registry names to allow for easier development
* Authenticate with balena registry
* Allow barys to build multiple bitbake targets

### DIFF
--- a/automation/jenkins_build-containers.sh
+++ b/automation/jenkins_build-containers.sh
@@ -5,6 +5,7 @@ set -ev
 DOCKERFILES=( Dockerfile_yocto-build-env Dockerfile_balena-push-env )
 
 REVISION=$(git rev-parse --short HEAD)
+NAMESPACE=${NAMESPACE:-resin}
 
 # Get the absolute script location
 pushd `dirname $0` > /dev/null 2>&1
@@ -20,12 +21,12 @@ for DOCKERFILE in "${DOCKERFILES[@]}"
 do
   REPO_NAME=${DOCKERFILE#"Dockerfile_"}
   # Build
-  docker build --pull --no-cache --tag resin/${REPO_NAME}:${REVISION} -f ${SCRIPTPATH}/${DOCKERFILE} ${SCRIPTPATH}
+  docker build --pull --no-cache --tag ${NAMESPACE}/${REPO_NAME}:${REVISION} -f ${SCRIPTPATH}/${DOCKERFILE} ${SCRIPTPATH}
 
   # Tag
-  docker tag resin/${REPO_NAME}:${REVISION} resin/${REPO_NAME}:latest
+  docker tag ${NAMESPACE}/${REPO_NAME}:${REVISION} ${NAMESPACE}/${REPO_NAME}:latest
 
   # Push
-  docker push resin/${REPO_NAME}:${REVISION}
-  docker push resin/${REPO_NAME}:latest
+  docker push ${NAMESPACE}/${REPO_NAME}:${REVISION}
+  docker push ${NAMESPACE}/${REPO_NAME}:latest
 done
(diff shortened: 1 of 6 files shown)


automation/jenkins_build-containers.sh
```diff
@@ -5,6 +5,7 @@ set -ev
 DOCKERFILES=( Dockerfile_yocto-build-env Dockerfile_balena-push-env )
 
 REVISION=$(git rev-parse --short HEAD)
+NAMESPACE=${NAMESPACE:-resin}
```
mtoman, line 8: Will `NAMESPACE` be set from the outside? As a part of `s,resin,balena,` we could make it default to `balena` and have the caller overwrite it to `resin`.

automation/jenkins_build.sh
```diff
@@ -3,6 +3,7 @@
 set -ex
 
 BUILD_CONTAINER_NAME=yocto-build-$$
+NAMESPACE=${NAMESPACE:-resin}
```
mtoman, line 6: Same as above

automation/prepare-and-start.sh
```diff
@@ -25,6 +25,21 @@ echo "[INFO] Starting docker."
...
+		echo "[INFO] Logging into $DEPLOY_TO as ${BALENAOS_ACCOUNT}"
+		export BALENARC_BALENA_URL=balena-staging.com
+		balena login --token ${BALENAOS_STAGING_TOKEN}
+	fi
+else
+	if [ -n "${BALENAOS_PRODUCTION_TOKEN}" ]; then
+		echo "[INFO] Logging into $DEPLOY_TO as ${BALENAOS_ACCOUNT}"
+		balena login --token ${BALENAOS_PRODUCTION_TOKEN}
+	fi
+fi
```
mtoman, line 41: This seems too complicated, can't we just have a single `BALENAOS_TOKEN` and overwrite the URL when `"$DEPLOY_TO" = "staging"`? Or is this something we have today and need to be compatible with it?

automation/jenkins_build-containers.sh
alexgg: @mtoman currently we push to https://hub.docker.com/r/resin/resinos, and there are no plans to change this to balena as we are moving away from pushing to dockerhub. So the default is "resin" not to  break things. In the short term this will change to use "balena deploy" instead of "docker push", but at least this change allows to set a NAMESPACE for a personal dockerhub account so that as developers we can test things without pushing to a production repository.

automation/jenkins_build.sh
alexgg: @mtoman answered above too.

automation/prepare-and-start.sh
alexgg: @mtoman you're right that it can be simplified. 
```
# Authenticate with Balena registry if required                                    
BALENAOS_ACCOUNT="balena_os"                                                       
BALENAOS_TOKEN=${BALENAOS_PRODUCTION_TOKEN}                                        
if [ "$DEPLOY_TO" = "staging" ]; then                                              
         BALENAOS_TOKEN=${BALENAOS_STAGING_TOKEN}                                   
         export BALENARC_BALENA_URL=balena-staging.com                              
fi                                                                                 
if [ -n "${BALENAOS_TOKEN}" ]; then                                                
         echo "[INFO] Logging into $DEPLOY_TO as ${BALENAOS_ACCOUNT}"               
        balena login --token ${BALENAOS_TOKEN}                                     
fi  
```
Moving forward we will be pulling blocks from the balena registry, so authenticating here is credentials are provided prepares us for that.

alexgg: Fixed, could you please re-review @mtoman ?

alexgg:
Addressed all review comments so I will self certify.

alexgg:
@balena-ci I self-certify!

automation/jenkins_build-containers.sh
mtoman: makes sense, thanks

PR #187, opened 2021-02-25
Dynamically build helper containers

alexgg:
Instead of using the master tag for helper containers, use those with a tag that matches the repository revision building them if required.

This allows to make changes to the helper containers without affecting the build of previous releases.

### DIFF
--- a/automation/balena-lib.sh
+++ b/automation/balena-lib.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+script_dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+BALENA_YOCTO_SCRIPTS_REVISION=$(cd "${script_dir}" && git rev-parse --short HEAD)
+
+# Pull a helper image building a matching version if required
+# Inputs:
+# $1: Dockerfile name
+docker_pull_helper_image() {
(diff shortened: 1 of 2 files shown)


automation/balena-lib.sh
```diff
@@ -0,0 +1,30 @@
...
+BALENA_YOCTO_SCRIPTS_REVISION=$(cd "${script_dir}" && git rev-parse --short HEAD)
+
+# Pull a helper image building a matching version if required
+# Inputs:
+# $1: Dockerfile name
+docker_pull_helper_image() {
+	local _dockerfile_name="$1"
+	local _image_name=""
+	local _image_prefix=""
+	_image_name="${_dockerfile_name%".template"}"
```
klutchell, line 13: When does this `template` suffix come into play? I guess when we want to make helper containers that are machine-specific but we don't have any examples yet?

alexgg: @klutchell they'll come into play with the changes to build blocks (the artist formerly know as hostapp extensions). This actually comes from me splitting up a previous PR. The other part, which I still need to work on to cater for the nre changes, is https://github.com/balena-os/balena-yocto-scripts/tree/alexgg/build-hostext.

klutchell (approved):
LGTM

PR #189, opened 2021-02-25
Use version controlled helper containers

alexgg:
This set of commits move the helper containers used to build and push releases to be source controlled instead of using the `latest` version.

This means that from this version onwards we will be able to modify helper scripts and they will only affect releases that use the balena-yocto-script revision that contains the changes.

Note that older revisions that use master will still break is `master` is updated to something that is not working. Those older revisions should be updated to this revision.


alexgg:
@balena-ci I self-certify!

PR #190, opened 2021-02-26
Dockerfile_yocto-build-env: Install required python3-distutils module

floion:
Otherwise bitbake fails with the following error:

Your Python 3 is not a full install. Please install the module
distutils.sysconfig (see the Getting Started guide for further
information).

Change-type: patch


floion:
@balena-ci I self-certify!

PR #192, opened 2021-03-03
Dockerfile_yocto-build-env: Install required git-lfs and bison modules

vicgal:
Otherwise bitbake fails with the following error:

The following required tools (as specified by HOSTTOOLS) appear
to be unavailable in PATH, please install them in order to proceed
  git-lfs bison

Change-type: patch
Signed-off-by: Vicentiu Galanopulo <vicentiu@balena.io>


alexgg:
`git-lfs` or  `bison` no longer appear in the `HOSTTOOLS` defined in the `local.conf.sample` file for any supported device type, closing this.

PR #198, opened 2021-03-24
Splitting functionality from jenkins_build to script libraries so they can be reused

alexgg:
Specifically there are:

* balena-build.sh: This script can be used standalone to call a containerized barys build. Helpful for customers that don't want to setup an OS and to keep a the same shared state between different machines.
* The `balena-deploy.inc` file contains a `balena_deploy_hostapp` function that deploys a hostapp.


automation/balena-lib.inc
```diff
@@ -28,3 +46,97 @@ balena_lib_docker_pull_helper_image() {
...
+	BALENARC_BALENA_URL=${_api_env} balena login --token "${_token}"
+}
+
+# Login to dockerhub
+balena_lib_dockerhub_login() {
+    local _user="${1}"
+    local _pwd="${2}"
+
+    if [ -z "${_user}" ]; then
+        if [ -n "${JOB_NAME}" ]; then
```
klutchell, line 128: I can't find where `JOB_NAME` is actually used. Is it a legacy variable we should clean up?

automation/jenkins_build-containers.sh
```diff
@@ -8,16 +8,9 @@ SCRIPTPATH=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
...
+source "${SCRIPTPATH}/balena-lib.inc"
 
-if [ -z "${JOB_NAME}" ]; then
-    echo "[ERROR] No job name specified."
-    exit 1
-fi
-
-echo "Login to docker as ${DOCKERHUB_USER}"
-docker login -u "${DOCKERHUB_USER}" -p "${DOCKERHUB_PWD}"
+balena_lib_dockerhub_login
```
klutchell, line 13: I guess in this case we are expecting DOCKERHUB_USER and DOCKERHUB_PWD to be set in the env so we aren't passing args?

automation/balena-lib.inc
alexgg: @klutchell it's set by Jenkins automatically. Here it is used to distinguish whether we are called as part of a Jenkins build so the credentials are expected to be in the environment, or not, in which case the login will ask the user for credentials (this is for local builds)

automation/jenkins_build-containers.sh
alexgg: @klutchell if it's a Jenkins build then yes, if not "docker login" will either used cached credentials or prompt the user.

automation/balena-lib.inc
klutchell: Ah, of course, I forgot that we use JOB_NAME to determine if it's Jenkins.

alexgg:
@balena-ci rebase

alexgg:
Tested with a raspberrypi4-64 build & staging deployment and resolved all the review comments.

alexgg:
@balena-ci I self-certify!

PR #203, opened 2021-03-30
Organize include files and entry scripts into directories

alexgg:
This makes the script easier to identify and use. It also sets a good base to increase the number of scripts without having a messier repository.


alexgg:
Just file re-organization, tested on a test build & deploy raspberrypi4-64 job.

alexgg:
@balena-ci I self-certify!

PR #204, opened 2021-04-12
Move include files and entry scripts into its own folders, and miscellaneous fixes

alexgg:
Hopefully the last batch of refactoring.


alexgg:
@balena-ci I self-certify!

PR #206, opened 2021-04-21
Miscellaneous fixes

alexgg:
Some general fixes. These have been tested while working on the hostos blocks building.


alexgg:
@balena-ci I self-certify!

PR #210, opened 2021-05-06
Remove docker as a build dependency (pending on aufs deprecation)

alexgg:
This removes docker from the container helpers.


alexgg:
Won't do

PR #211, opened 2021-05-07
Generate config.js from test suites in meta-balena

klutchell:
For each test suite in meta-balena we need a a blob
of javascript added to the config.js manifest for leviathan.

Also allow overriding the default config values with a
suite.env file in the root of the suite directory.

Example: _tests/suites/os/suites.env_

```
NETWORK_WIRED=false
NETWORK_WIRELESS=true
DOWNLOAD_TYPE=local
INTERACTIVE_TESTS=false
IMAGE=my-image.img.gz
API_URL=balena-cloud.com
BALENA_APPLICATION=testbot-rig
```

Having an `.env` file seemed the most reliable way for bash to generate a valid `.js` manifest without also having to parse a `.js` file. If these were in pure JSON I would just use `jq`.

Change-type: patch
Changelog-entry: Generate config.js from test suites in meta-balena

### DIFF
--- a/automation/include/balena-deploy.inc
+++ b/automation/include/balena-deploy.inc
@@ -66,6 +66,39 @@ balena_deploy_hostapp() {
 	balena_lib_docker_remove_helper_images "balena-push-env"
 }
 
+add_test_suite () {
+	local config_js="${1}"
+	local suite_path="${2}"
+
+	# skip this dir if a suite.js is not present
+	[ -f "${suite_path}/suite.js" ] || return
+
+	# source suite.env if present to allow overrides
+	# shellcheck disable=SC1091
+	. "${suite_path}/suite.env" || true
+
+	# write the javascript blob assuming there will be another element to follow
+	# if not the final close/open brackets will be replaced with the array closure
+	cat << EOF >> "${config_js}"
+	deviceType: \`${JOB_NAME#yocto-}\`,
+	suite: \`\${__dirname}/../suites/$(basename "${suite_path}")\`,
+	config: {
+		networkWired: ${NETWORK_WIRED:-false},
+		networkWireless: ${NETWORK_WIRELESS:-true},
+		downloadType: '${DOWNLOAD_TYPE:-local}',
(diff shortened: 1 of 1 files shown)


automation/include/balena-deploy.inc
```diff
@@ -66,6 +66,39 @@ balena_deploy_hostapp() {
 	balena_lib_docker_remove_helper_images "balena-push-env"
 }
 
+add_test_suite () {
+	local config_js="${1}"
+	local suite_path="${2}"
+
+	# skip this dir if a suite.js is not present
+	[ -f "${suite_path}/suite.js" ] || return
```
robertgzr, line 74: with the `set -e` at the top, won't this make it exit?

rcooke-warwick:
@klutchell , the upcoming tests I;m adding require a config.json to have a config attribute like this (apikey and org):

```js
config: {
        networkWired: false,
        networkWireless: true,
        downloadType: 'local',
        interactiveTests: false, // redundant
        balenaApiKey: process.env.API_KEY,
        balenaApiUrl: 'balena-cloud.com',
        organization: process.env.ORG
    },
   ```

**Comment by klutchell** [comment]
> @klutchell , the upcoming tests I;m adding require a config.json to have a config attribute like this (apikey and org):

Awesome! That's so much cleaner, I can start writing for that instead and update this PR accordingly. Should the `config.json` be in the root of the suite? It would just be strings and booleans like this, correct?

Example: _tests/suites/os/config.json_
```json
{
        "networkWired": false,
        "networkWireless": true,
        "downloadType": "local",
        "interactiveTests": false,
        "balenaApiKey": "process.env.API_KEY",
        "balenaApiUrl": "balena-cloud.com",
        "organization": "process.env.ORG"
}
```

alexgg (approved):
lgtm

rcooke-warwick:
@klutchell it  should be just strings and booleans yeah

klutchell:
Closing in favour of a pre-defined Leviathan job configuration as included in this PR.
- https://github.com/balena-os/meta-balena/pull/2174

PR #212, opened 2021-05-10
Substitute device type in config.js if present

klutchell:
Connects-to: https://github.com/balena-os/meta-balena/pull/2174
Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #216, opened 2021-05-17
Allow access to OS developers to public apps

alexgg:
When creating a public app, set the access rights so that OS developers can see the app in their accounts.


alexgg:
@balena-ci I self-certify!

PR #219, opened 2021-05-18
Set docker version to match build system and do not force dockerd settings

alexgg:
These will solve the container networking problems with the default bridge network in the build.


alexgg:
@balena-ci I self-certify!

PR #220, opened 2021-05-18
Rebrand resin to balena and use release_version to specify supervisor

alexgg:
* Barys and `jenkins_build.sh` argument `--supervisor-tag` becomes `--supervisor-release` in preparation for the OS to use `release_version` to fetch the supervisor image
* Build jobs needs to change the configuration variable `supervisorTag` with `supervisorRelease`
* Build jobs needs to change `metaResinBranch` to `metaBalenaBranch`

### DIFF
--- a/build/barys
+++ b/build/barys
@@ -22,7 +22,7 @@ BUILD_DIR=build
 REMOVEBUILD=no
 COMPRESS=no
 DEVELOPMENT_IMAGE=no
-SUPERVISOR_TAG=
+SUPERVISOR_VERSION=
 SHARED_DOWNLOADS=
 SHARED_SSTATE=
 DRY_RUN=no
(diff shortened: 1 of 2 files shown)


build/barys
```diff
@@ -22,7 +22,7 @@ BUILD_DIR=build
 REMOVEBUILD=no
 COMPRESS=no
 DEVELOPMENT_IMAGE=no
-SUPERVISOR_TAG=
+SUPERVISOR_VERSION=
```
klutchell, line 25: Any specific reasoning for using `VERSION` in barys vs `Release` in jenkins_build.sh? It's all the same to me but it could be aligned?

alexgg: No reason, I'll align them.

alexgg:
@balena-ci I self-certify!

PR #222, opened 2021-05-24
versionbot: Add changelog yml file

alexgg:
This file allows other components to uniquely parse the information that
is contained in the changelog. It will be automatically managed by
versionist by appending the new commits on top. This is needed to
provide nested-changelogs.

Change-type: patch
Changelog-entry: Add a parsable representation of the changelog


alexgg:
@balena-ci I self-certify!

PR #223, opened 2021-05-25
Add missing library functions to successfully build hostOS blocks 

alexgg:
Extend script libraries to support hostOS blocks building.


alexgg:
@balena-ci I self-certify!

PR #224, opened 2021-06-01
Miscellaneous fixes/improvements

alexgg:
* Add a conditional verbosity mode
* Fix installation directory for block deployment scripts
* Modify block contracts to specify separately recipes and packages
* Fix setting app roles on non-production environments


alexgg:
@balena-ci I self-certify!

PR #226, opened 2021-06-16
balena-deploy-block: source balena-lib script library

alexgg:
Some functions in balena-api require this library.

Change-type: patch


alexgg:
@balena-ci I self-certify!

PR #227, opened 2021-06-18
balena-deploy-block: Do not add variant suffix to release tag

alexgg:
Fix the legacy way of tagging releases with a string variant and the OS
version as this is still being used by the UI.

Change-type: patch


alexgg:
@balena-ci I self-certify!

PR #228, opened 2021-06-18
balena-deploy: Provide working directory when deploying hostapp

alexgg:
When creating a public app, the workdir is used to retrieve the URL
used in the public app creation.

Change-type: patch


alexgg:
@balena-ci I self-certify!

PR #230, opened 2021-06-29
balena-lib: Fix device dir when running in Jenkins containers

klutchell:
Avoid returning an empty path when we are running in containers
where /work does not exist.

Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #235, opened 2021-09-28
barys: Remove RESINHUP setting

alexgg:
This has now been removed from meta-balena.

Change-type: patch


alexgg:
@balena-ci I self-certify!

PR #236, opened 2021-10-01
jenkins_generate_ami.sh: Add missing argument to balena_lib_get_dt_arch

mtoman:


mtoman:
@balena-ci I self-certify!

PR #237, opened 2021-10-01
Update balena-cli to 12.50.1 in automation dockerfiles

mtoman:
This is necessary for AMI preloading to work, additionally it has been more than a year since the last update, we should keep up.


mtoman:
@balena-ci I self-certify!

PR #238, opened 2021-10-01
balena-generate-ami.sh: correctly mount boot partition when injecting SSH key

mtoman:


mtoman:
@balena-ci I self-certify!

PR #248, opened 2021-11-04
Automate labelling of ESR hostapps

alexgg:
ESR hostapps need the following labels:

* release-policy: esr
* esr-current
* esr-next
* esr-sunset

This set labels them appropriately.


alexgg:
@balena-ci I self-certify!

PR #250, opened 2021-11-12
Use proper semver for meta-balena-base ESR hostapp labels

alexgg:


alexgg:
@balena-ci I self-certify!

PR #255, opened 2022-01-14
prepare-and-start: Do not ignore token when passed in command line

alexgg:
The balena-build.sh script accepts a token in the command line. This
commit modifies the prepare-and-start.sh script to use it if
available.

Change-type: patch


alexgg:
@balena-ci I self-certify!

alexgg:
@balena-ci I self-certify!

PR #257, opened 2022-01-17
balena-deploy-block: Check for existing tag only in specific hostapp

alexgg:
The same version tag can exist across different hostapps.

Change-type: patch


alexgg:
@balena-ci rebase

alexgg:
@balena-ci I self-certify!

PR #258, opened 2022-01-27
balena-deploy.inc: Do no deploy device logo to deprecated endpoint

floion:
As per https://jel.ly.fish/a533a6bd-1c98-412c-ae60-8427e3f7b005,
the /device-types/v1 endpoint is deprecated and we can remove
the logo from the device repos so we can also stop handling it
from here. Instead, the logo will be used from the hw contracts.

Change-type: patch


floion:
@balena-ci I self-certify!

PR #261, opened 2022-03-02
jenkins_build: Use recurse-submodules when checking out meta-balena

klutchell:
If meta-balena has submodule updates we want to make sure to use them.

Change-type: patch
See: https://github.com/balena-os/meta-balena/pull/2514


klutchell:
@balena-ci I self-certify!

PR #271, opened 2022-03-31
Bump balena-cli to v.Latest

ab77:
Change-type: minor


ab77:
@balena-ci I self-certify!

ab77:
@balena-ci I self-certify!

PR #274, opened 2022-04-14
deploy: ensure deployRawArtifact is empty instead of null

jakogut:
jq returns null by default when a given key isn't found, ensure that
when getting the value of deployRawArtifact, we get an empty variable
instead, which is checked later on to determine if that file should be
deployed

Change-type: patch
Signed-off-by: Joseph Kogut <joseph@balena.io>


alexgg (approved):
lgtm

alexgg:
@jakogut maybe PR one of the failing device types on this version to make sure it builds fine before the merge gets spread to all repositories.

PR #276, opened 2022-04-26
balena-lib: fix 445d6d1fcfce97f85ffcfedc0083eb658a734321

floion:
Commit 445d6d1fcfce97f85ffcfedc0083eb658a734321 does not fix
the issue because it is missing the --global arg without which
the git config command fails.

Change-type: patch


floion:
@balena-ci I self-certify!

PR #280, opened 2022-05-31
balena-lib: Tag ESR fleets with patch independent tags

alexgg:
Otherwise patch updates of ESR branches move the ESR phase when they
should not. For example, if 2022.1.1 is current, 2022.1.2 is also
current and should not move 2022.1.1 to sunset.

Change-type: patch


alexgg:
Verifying in https://jenkins.product-os.io/job/yocto-raspberrypi4-64/3166/console

alexgg:
@balena-ci I self-certify!

alexgg:
@balena-ci republish

PR #281, opened 2022-06-01
balena-lib: release_finalize: Do not retag phase for ESR branch patch

alexgg:
When we patch an ESR branch, for example from v2022.1.0 to v2022.1.1,
do not update the next, current, sunset ESR  phases as they remain the
same.

Change-type: patch


alexgg:
@balena-ci I self-certify!

PR #284, opened 2022-06-28
balena-deploy: s/resin-img/balena-img

jakogut:
As part of rebranding, resin docker repos were renamed to balena, and
resin/resin-img no longer receives updates. Change the image we pull to
process OS images from resin/resin-img to balena/balena-img.

Change-type: patch
Signed-off-by: Joseph Kogut <joseph@balena.io>

### DIFF
--- a/automation/include/balena-deploy.inc
+++ b/automation/include/balena-deploy.inc
@@ -270,7 +270,7 @@ balena_deploy_to_s3() {
 
 	local _s3_cmd="s4cmd --access-key=${_s3_access_key} --secret-key=${_s3_secret_key}"
 	local _s3_sync_opts="--recursive --API-ACL=${_s3_policy}"
-	docker pull ${_namespace}/balena-img:master
+	docker pull "${_namespace}/balena-img:4.1.2"
 	docker run --rm -t \
 		-e BASE_DIR=/host/images \
 		-e S3_CMD="$_s3_cmd" \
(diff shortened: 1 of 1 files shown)


automation/include/balena-deploy.inc
```diff
@@ -270,7 +270,7 @@ balena_deploy_to_s3() {
 
 	local _s3_cmd="s4cmd --access-key=${_s3_access_key} --secret-key=${_s3_secret_key}"
 	local _s3_sync_opts="--recursive --API-ACL=${_s3_policy}"
-	docker pull ${_namespace}/resin-img:master
+	docker pull ${_namespace}/balena-img:master
```
alexgg, line 273: Are balena-img deployments tagged? Should we fix a tag instead of pulling master to avoid upstream changes breaking OS deployments?

alexgg (commented):
lgtm, just a comment about versions

automation/include/balena-deploy.inc
alexgg: @jakogut I proposed fixing a tag so that if a specific OS version works (which fixes balena-yocto-script to a sha1), it never breaks.
My first instinct is that if there is some new feature the OS needs from a newer balena-img, the right thing is to PR and explain the update in the commit.
Not sure what's the actual use of overriding this in Jenkins, do you have a use case in mind?

jakogut: > Not sure what's the actual use of overriding this in Jenkins, do you have a use case in mind?

Nope, that reasoning makes sense to me. I was thinking it's one less thing we need to make PRs for, but your explanation makes sense, I just hadn't thought about it that way. I'll update this with a tag.

jakogut:
@balena-ci I self-certify!

PR #288, opened 2022-07-25
balena-api: On hostapp creation, set class to app

alexgg:
Applications can be fleet, blocks or apps, and hostapps need to be set
as apps.

Change-type: patch


alexgg:
@balena-ci I self-certify!

PR #289, opened 2022-08-04
barys: Add balena-rust to syntax conversion step

alexgg:
Recent versions of meta-balena include a balena-rust layer used to
specify a distro-set Rust version across all supported Yocto versions

As such, the syntax of this layer also needs to be converted.

Relates-to: https://github.com/balena-os/meta-balena/pull/2750
Change-type: patch


alexgg:
@balena-ci I self-certify!

PR #291, opened 2022-11-10
jenkins_generate_ami: use a different token

alexgg:
The AMI preload application no longer lives in the balena_os organization so allow to provide a different token.

Change-type: patch


alexgg:
@balena-ci I self-certify!

PR #295, opened 2022-12-19
jenkins_build: secure boot & FDE AMI images support

alexgg:
Building OS development images is useful for example when debugging and testing AMI images.

Change-type: patch


alexgg:
@balena-ci I self-certify!

PR #297, opened 2022-12-27
test AMIs before making them public

alexgg:


alexgg:
Tested in AWS playground

ab77 (approved):

PR #300, opened 2023-01-19
balene-generate-ami: remove key name

alexgg:
No need to provide a key name as the instance only needs to be accessible via balenaCloud.

Change-type: patch


alexgg:
@balena-ci I self-certify!

PR #303, opened 2023-02-28
balena-api: ignore quotes from API replies

alexgg:
The balenaAPI has replied both `OK` and `"OK"` to patch requests. Accept any of them as long as they contain OK.

Change-type: patch


alexgg:
lgtm

PR #306, opened 2023-04-21
barys: fix removal of equals sign from argval

jakogut:
When parsing additional variables to be passed to the bitbake build, keys and values are split using equals as a delimiter. However, the splitting process does not split only on the first occurrence, which results in removing equals signs from the value as well. This is problematic with base64 encoded strings, which are padded with equals signs.

Split only on the first occurrence, leaving the value intact.

Change-type: patch


jakogut:
lgtm

PR #308, opened 2023-05-10
balena-generate-ami: comply with default AMI quotas

alexgg:
There is a default limit of 5 public AMI images per region. Make the oldest public image back to private before publishing a new image.

Change-type: patch


alexgg:
Checks:
- [] Running machines that used a public AMI are unaffected when the AMI is made private

alexgg:
Rebased into https://github.com/balena-os/balena-yocto-scripts/pull/310

PR #309, opened 2023-05-16
balena-deploy: add secureBoot feature into contract

alexgg:
When building signed images, add the secureBoot feature flag into the OS contract. This is needed for other components to identify secureBoot compatible software releases.

Change-type: patch


alexgg:
lgtm

PR #310, opened 2023-05-23
Secure boot support

alexgg:
* Add the secureboot feature to OS contracts
* For signed images, opt in secure boot in the AMI images


alexgg:
lgtm

PR #312, opened 2023-06-15
balena-deploy: Newer meta-balena do not create a kernel_source tarball

alexgg:
This artifact did not contain kernel sources since Yocto Thud.

Changelog-entry: The kernel_source tarball is no longer created by meta-balena


alexgg:
lgtm

PR #316, opened 2023-09-27
Fail deploy if script fails to find DT

rcooke-warwick:
There is a bug when the curl GET to check if a DT is private or not fails - it assumes the device type is private. This can lead to s3 artifacts for public DT's not being marked as public-read, if the curl fails (for example if there are api errors or connectivity issues)


rcooke-warwick:
@alexgg do you think this would have an unintended side-effect when there is a device type deployed for the first time? I'm not sure what order things happen in, if the device type only appears as a supported DT in the api before or after the first deploy?

alexgg:
> @alexgg do you think this would have an unintended side-effect when there is a device type deployed for the first time? I'm not sure what order things happen in, if the device type only appears as a supported DT in the api before or after the first deploy?

hey @rcooke-warwick the device type appears in the API with the contract PR, the deployment only creates the hostapp fleet and the S3 folder. I think it should be fine.

alexgg (approved):

PR #324, opened 2024-01-30
Build and publish helper images with Flowzone

klutchell:
Rather than build helper images on demand, we will publish them to ghcr.io with every revision of this project.

Change-type: minor


klutchell:
Integration tests: https://github.com/balena-os/balena-generic/pull/405

automation/entry_scripts/balena-build-block.sh
alexgg, line 1: @klutchell retesting all these changes is very time consuming - what I would suggest it to remove the paths that are not used (and probably never will), For example the whole `jenkins_build-blocks.sh` -> balena_build_block()  in `automation/include/balena-deploy.inc` and `balena-build-block.sh` is unused I believe.

I would PR that cleanup first, and then redo this PR to avoid having to re-implement those.

alexgg (changes requested):
lgtm - I could approve as is. Just a note that some of the changes are on paths that are not currently used to it would maybe simplify things to PR the clean-up and then modify this PR.

But anyway I could approve it as it too as the paths are unused anyway - feel free to self certify.

alexgg (approved):

PR #327, opened 2024-02-07
automation/include: Pass helper image version

floion:
This is needed by the balena-push-env helper image that needs the balena-yocto-scripts version but has no way of getting that info otherwise.

Change-type: patch


acostach:
lgtm

PR #328, opened 2024-02-10
Enable S3 Server Side Encryption flags

klutchell:
Change-type: patch


acostach:
Tested with un-finalized releases in staging and production. Internal thread https://balena.zulipchat.com/#narrow/stream/345889-balena-io.2Fos/topic/yocto-build-env/near/420891565

acostach:
lgtm

PR #329, opened 2024-02-13
Create workflow to build and deploy balenaOS

klutchell:


.github/workflows/dispatch-yocto-build-deploy.yml
```diff
@@ -0,0 +1,124 @@
...
+    types: [opened, synchronize]
+    branches:
+      - "main"
+      - "master"
+
+  workflow_dispatch:
+    # you may only define up to 10 `inputs` for a `workflow_dispatch` event
+    inputs:
+      device-repo:
+        description: balenaOS device repository (owner/repo)
```
alexgg, line 19: Thinking whether we should just place the build script in the device repository (maybe an include from a common location) so that this information is provided by github.

.github/workflows/dispatch-yocto-build-deploy.yml
```diff
@@ -0,0 +1,124 @@
...
+        default: master
+      meta-balena-ref:
+        description: meta-balena ref if not the currently pinned version
+        required: false
+        type: string
+      yocto-scripts-ref:
+        description: balena-yocto-scripts ref if not the currently pinned version
+        required: false
+        type: string
+      machine:
```
alexgg, line 36: Again, this is something that we should probably just extract from the device repository.

.github/workflows/dispatch-yocto-build-deploy.yml
```diff
@@ -0,0 +1,124 @@
...
+        description: yocto board name
+        required: true
+        type: string
+        default: generic-amd64
+      environment:
```
alexgg, line 41: We need to take into account how this will change once deploys work over the API.

.github/workflows/dispatch-yocto-build-deploy.yml
```diff
@@ -0,0 +1,124 @@
...
+        description: Select deploy environment
+        required: false
+        type: choice
+        options:
+          - >
+            {
+              "environment": "balena-staging.com",
+              "s3-bucket": "resin-staging-img",
+              "s3-region": "us-east-1",
+              "aws-subnet": "subnet-0d73c1f0da85add17",
```
alexgg, line 51: Are subnet and security group needed to deploy to S3?

.github/workflows/dispatch-yocto-build-deploy.yml
```diff
@@ -0,0 +1,124 @@
...
+            }
+          - >
+            {
+              "environment": "balena-cloud.com",
+              "s3-bucket": "resin-production-img-cloudformation",
+              "s3-region": "us-east-1",
+              "aws-subnet": "subnet-02d18a08ea4058574",
+              "aws-security-group": "sg-057937f4d89d9d51c"
+            }
+      deploy-s3:
```
alexgg, line 62: How do these interact with the global `deploy`? Could we make it to have just the `deploy` yes/no switch, and avoid all these s3/hostapp/ami switches?

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,639 @@
...
+      BALENA_API_KEY:
+        description: balena API key for deploying releases
+        required: false
+      S3_ACCESS_KEY:
+        description: S3 access key
+        required: false
+      S3_SECRET_KEY:
+        description: S3 secret key
+        required: false
+      DOCKERHUB_USER:
```
alexgg, line 15: Let's remove the dockerhub credentials, we don't use them anymore and we won't backport the new build script to old releases anyway.

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,639 @@
...
+        description: Dockerhub user for pulling private helper images
+        required: false
+      DOCKERHUB_TOKEN:
+        description: Dockerhub token for pulling private helper images
+        required: false
+      SIGN_KMOD_KEY_APPEND:
+        description: Base64-encoded public key of a kernel module signing keypair
+        required: false
+      SIGN_API_KEY:
```
alexgg, line 24: The description is wrong here. This is the balenaCloud API key that gives access to the signing server. Nothing to do with module signing, those only need to include the public key above.

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,639 @@
...
+      meta_balena_version: ${{ steps.build-lib.outputs.meta_balena_version }}
+      yocto_scripts_ref: ${{ steps.build-lib.outputs.yocto_scripts_ref }}
+
+    defaults:
+      run:
+        working-directory: .
+        shell: bash --noprofile --norc -eo pipefail -x {0}
+
+    steps:
+      # https://github.com/actions/checkout
```
alexgg, line 166: We can probably avoid this clone if the build script lives with the device repository.

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,639 @@
...
+      - name: Update balena-yocto-scripts submodule to ${{ inputs.yocto-scripts-ref }}
+        if: inputs.yocto-scripts-ref != ''
+        working-directory: ./balena-yocto-scripts
+        run: |
+          git config --add remote.origin.fetch '+refs/pull/*:refs/remotes/origin/pr/*'
+          git fetch --all
+          git checkout --force "${{ inputs.yocto-scripts-ref }}"
+          git submodule update --init --recursive
+
+      - name: Device repository check
```
alexgg, line 193: This should come before updating meta-balena and balena-yocto-scripts as those will only exist on a device repository.

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,639 @@
...
+
+    defaults:
+      run:
+        working-directory: .
+        shell: bash --noprofile --norc -eo pipefail -x {0}
+
+    steps:
+      # TODO: clone balena-yocto-scripts @ yocto_scripts_ref instead
+      # https://github.com/actions/checkout
+      - name: Clone device repository
```
alexgg, line 303: Again, if the build script is part of the device repository we should not need to do this clone again.

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,639 @@
...
+        uses: actions/checkout@v3
+        with:
+          repository: ${{ inputs.device-repo || github.repository }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+          ref: ${{ inputs.device-repo-ref || github.ref }}
+          submodules: true
+
+      - name: Update balena-yocto-scripts submodule to ${{ inputs.yocto-scripts-ref }}
```
alexgg, line 311: We could also avoid this update as we have done it before.

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,639 @@
...
+        if: inputs.yocto-scripts-ref != ''
+        working-directory: ./balena-yocto-scripts
+        run: |
+          git config --add remote.origin.fetch '+refs/pull/*:refs/remotes/origin/pr/*'
+          git fetch --all
+          git checkout --force "${{ inputs.yocto-scripts-ref }}"
+          git submodule update --init --recursive
+
+      # https://github.com/actions/download-artifact
```
alexgg, line 320: Don't these persist in the build node? Re-downloading them when all of these happens in the same machine is not optimal, specially with big artifacts.

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,639 @@
...
+          s3_bucket_path="${{ inputs.s3-bucket}}/images"
+          if [ "${{ inputs.deploy-esr}}" = "true" ]; then
+            s3_bucket_path="${{ inputs.s3-bucket}}/esr-images"
+          fi
+
+      # https://github.com/docker/setup-buildx-action
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      # login required to pull private balena/balena-img image
```
alexgg, line 352: why is balena/balena-img private?

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,639 @@
...
+      # https://github.com/docker/login-action
+      - name: Login to Docker Hub
+        uses: docker/login-action@343f7c4344506bcbf9b4de18042ae17996df046d # v3.0.0
+        with:
+          registry: docker.io
+          username: ${{ secrets.DOCKERHUB_USER }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      # https://github.com/docker/build-push-action
+      - name: Build helper image
```
alexgg, line 362: Could we not use a pre-built image instead of building for every build?

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,639 @@
...
+        uses: docker/build-push-action@v5.1.0
+        with:
+          context: "${{ github.workspace }}/balena-yocto-scripts/automation"
+          file: Dockerfile_s3-deploy-env
+          push: false
+          build-args: |
+            DEPLOYER_UID=${{ steps.s3-lib.outputs.deployer_uid }}
+            DEPLOYER_GID=${{ steps.s3-lib.outputs.deployer_gid }}
+
+      - name: Deploy to S3
```
alexgg, line 373: Maybe we should look for a s3 deploy action that is maintained externally. Also, I'd probably push for file uploads to be used instead of wasting time and energy on this.

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,639 @@
...
+      WORKSPACE: ${{ github.workspace }}
+      VERBOSE: verbose
+
+    defaults:
+      run:
+        working-directory: .
+        shell: bash --noprofile --norc -eo pipefail -x {0}
+
+    steps:
+      # TODO: clone balena-yocto-scripts @ yocto_scripts_ref instead
```
alexgg, line 430: Same concersn about re-cloning - we should opmitize.

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,639 @@
...
+        if: inputs.yocto-scripts-ref != ''
+        working-directory: ./balena-yocto-scripts
+        run: |
+          git config --add remote.origin.fetch '+refs/pull/*:refs/remotes/origin/pr/*'
+          git fetch --all
+          git checkout --force "${{ inputs.yocto-scripts-ref }}"
+          git submodule update --init --recursive
+
+      # https://github.com/actions/download-artifact
+      - name: Download build artifacts
```
alexgg, line 450: Same concerns about build artifacts - they are already on the build node.

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,639 @@
...
+      SIGN_API_URL: "${{ inputs.sign-api-url }}"
+
+    defaults:
+      run:
+        working-directory: .
+        shell: bash --noprofile --norc -eo pipefail -x {0}
+
+    steps:
+      # TODO: clone balena-yocto-scripts @ yocto_scripts_ref instead
+      # https://github.com/actions/checkout
```
alexgg, line 559: :)

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,639 @@
...
+            echo "SECURE_BOOT_FEATURE_FLAG=yes" >> $GITHUB_ENV
+          else
+            echo "SECURE_BOOT_FEATURE_FLAG=no" >> $GITHUB_ENV
+          fi
+
+      # https://github.com/docker/setup-buildx-action
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      # https://github.com/docker/build-push-action
```
alexgg, line 592: Prebuilt?

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,639 @@
...
+      - name: Build helper image
+        id: helper-image
+        uses: docker/build-push-action@v5.1.0
+        with:
+          context: "${{ github.workspace }}/balena-yocto-scripts/automation"
+          file: Dockerfile_yocto-build-env
+          target: yocto-build-env
+          push: false
+
+      - name: Deploy to balena
```
alexgg, line 602: Here is were the deploy to balena action would come in.

automation/Dockerfile_s3-deploy-env
```diff
@@ -0,0 +1,13 @@
+# private base image requires docker login
+FROM balena/balena-img:6.20.26
+
+ARG DEPLOYER_UID
+ARG DEPLOYER_GID
+
+# hadolint ignore=DL3008
```
alexgg, line 7: Using a standard S3 action would avoid maintaining this image.

alexgg (changes requested):
Probably a lot of things you already know and have decided to leave for later :)

.github/workflows/dispatch-yocto-build-deploy.yml
klutchell: This `dispatch-` workflow is an example of what will be in each device repo, and it is used to call the shared workflow like we do with Flowzone.

The device-repo field is only required when testing it from balena-yocto-scripts, and would not be exposed when this script is installed on device repos. In fact, on device repos we might only use the `pull_request` trigger and not allow manual execution via `workflow_dispatch` at all, instead using this copy in balena-yocto-scripts for manual execution and testing.

.github/workflows/dispatch-yocto-build-deploy.yml
klutchell: We will call the shared workflow once per machine that we want to build, so it will be hardcoded and not exposed unless we want to allow manual execution via the device repos directly, but that's a lot of parameter duplication.

.github/workflows/dispatch-yocto-build-deploy.yml
klutchell: They are required for the current publish AMI functions.

.github/workflows/dispatch-yocto-build-deploy.yml
klutchell: We still want to support both production and staging, with different keys and paths for both. The current plan is to use GitHub Environments to set these at the repo/org level and not need to provide them via these workflows at all.

See https://github.com/balena-os/.github/pull/8

.github/workflows/dispatch-yocto-build-deploy.yml
klutchell: The issue I see is that AMI is only deployed for a couple device types so it would be easy to set to true at the device-repo level when calling the shared workflow, and leave as false for everything else. At least until we can get the AMI from a source of truth like contracts. I'll add a TODO.

klutchell: Also being able to skip S3 and HostApp deploys is useful for testing at this stage. We may not always expose them.

.github/workflows/yocto-build-deploy.yml
klutchell: We need these to pull the private `balena/balenaimg` from DockerHub.

.github/workflows/yocto-build-deploy.yml
klutchell: ack, I'll update it

.github/workflows/yocto-build-deploy.yml
klutchell: All workflows need to perform a clone at the start, and since this is a shared workflow it will inherit the source repo and ref from the calling repository and not the inputs.

.github/workflows/yocto-build-deploy.yml
klutchell: Good call, I'll make the change.

.github/workflows/yocto-build-deploy.yml
klutchell: I've split the workflow into 4 jobs, and each job will select a new ephemeral runner with a fresh workspace.

This is why we need to upload/download artfiacts and re-clone at the start of each job.

The benefit is that the 3 deploy types can run in parallel after the build, hopefully saving time overall.

If it turns out to be faster to skip the multiple clones and artifact upload/download and keep it all in one job and workspace we can do that too, but I believe this will be faster overall.

.github/workflows/yocto-build-deploy.yml
klutchell: See above answer, we are using 4 separate jobs to make use of parallelism and each job is a fresh workspace.

If this turns out to not be faster, we can revert to one job and workspace.

.github/workflows/yocto-build-deploy.yml
klutchell: See above answer, we are using 4 separate jobs to make use of parallelism and each job is a fresh workspace.

If this turns out to not be faster, we can revert to one job and workspace.

.github/workflows/yocto-build-deploy.yml
klutchell: Yes, but then that image would have to be private as well if it is based on balena/balena-img. It's also something we could optimize later.

Also, there is a race condition that while this shared workflow currently lives with balena-yocto-scripts, changes to the helper images would not yet be published while this workflow is being tested. Right now this workflow runs on PRs to balena-yocto-scripts to make sure we haven't broken anything between the scripts and the workflow (deploys will be skipped). But if I change the content of a helper Docker image we would not be testing it unless I make sure this workflow runs after deploying the draft images. Something I can optimize in a future release for sure.

Not an issue for device repos that are using the master branch of balena-yocto-scripts, but still something that needs to be considered if the workflow will always live here.

.github/workflows/yocto-build-deploy.yml
klutchell: AFAIK the balena-s3 image actually handles the preparation of the deflate files and not just the uploading to S3. If that's not the case, then why did we bother using balena-s3 in the first place?

IIRC there is a bunch of special handling in balena-S3 for creating the required artifacts in the prepare.ts script before we start uploading.

.github/workflows/yocto-build-deploy.yml
klutchell: See above answer, we are using 4 separate jobs to make use of parallelism and each job is a fresh workspace.

If this turns out to not be faster, we can revert to one job and workspace.

.github/workflows/yocto-build-deploy.yml
klutchell: See above answer, we are using 4 separate jobs to make use of parallelism and each job is a fresh workspace.

If this turns out to not be faster, we can revert to one job and workspace.

.github/workflows/yocto-build-deploy.yml
klutchell: :)

.github/workflows/yocto-build-deploy.yml
klutchell: Yes, I would like to get to prebuilt images here, but as I mentioned in a previous answer when working on this repo directly, changes to the prebuilt images would not be published yet unless I set a specific order of operations.

It's only an issue when testing this workflow while also making changes to the helper images on this repo directly. Something we can workaround.

.github/workflows/yocto-build-deploy.yml
klutchell: Correct, will need a TODO here as well.

automation/Dockerfile_s3-deploy-env
klutchell: Se previous answer, I think prepare.ts prepares the deflate files for upload and it's not just generic S3 uploading.

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,663 @@
+name: "Yocto"
+
+on:
+  workflow_call:
+    secrets:
+      BALENA_API_KEY:
+        description: balena API key for deploying releases
+        required: false
+      S3_ACCESS_KEY:
```
ab77, line 9: @klutchell we should be using ephemeral AWS access here, not storing/passing tokens/keys around.

ab77 (changes requested):

.github/workflows/yocto-build-deploy.yml
klutchell: good call, we can look at how Flowzone does it

.github/workflows/dispatch-yocto-build-deploy.yml
```diff
@@ -0,0 +1,86 @@
...
+        description: Whether to deploy a hostApp container image to a balena environment
+        required: false
+        type: boolean
+        default: true
+      deploy-ami: # TODO: can we get this from a source of truth like contracts?
+        description: Whether to deploy an AMI to AWS
+        required: false
+        type: boolean
+        default: false
+      sign-image: # TODO: can we get this from a source of truth like contracts?
```
klutchell, line 52:
```suggestion
      sign-image:
```
we can remove this comment, as Alex explained that image signing is a feature of the build, and not specific to the device-type

so keep the input as-is, but remove the comment

rcooke-warwick:
In order to enable the correct s3 permissions for this workflow we need the following:
https://github.com/balena-io/environments-base/pull/72
https://github.com/balena-io/environment-staging/pull/3284

.github/workflows/yocto-build-deploy.yml
rcooke-warwick: setup of this AWS access:

https://github.com/balena-io/environments-base/pull/72
https://github.com/balena-io/environment-staging/pull/3284

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,659 @@
...
+      #   with:
+      #     provider: aws # default is linode
+      #     region: ${{ vars.AWS_REGION || vars.S3_REGION }}
+      #     access_key: ${{ secrets.AWS_ACCESS_KEY_ID || secrets.S3_SECRET_KEY }}
+      #     secret_key: ${{ secrets.AWS_SECRET_ACCESS_KEY || secrets.S3_ACCESS_KEY }}
+     
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: us-east-1 # Our role seems to be in us-east-1, this is a required parameter
```
ab77, line 450: its a require param for other reasons, IAM roles aren't region scoped

ab77:
```
  - configureAWSCredentials # https://github.com/aws-actions/configure-aws-credentials
    name: Configure AWS credentials
    id: aws_credentials
    uses: aws-actions/configure-aws-credentials@e3dd6a429d7300a6a4c196c26e071d42e0343502 # v4.0.2
    with:
      role-to-assume: ${{ inputs.aws_iam_role }}
      role-session-name: github-${{ github.job }}-${{ github.run_id }}-${{ github.run_attempt }}
      aws-region: ${{ inputs.aws_region }}
      # https://github.com/orgs/community/discussions/26636#discussioncomment-3252664
      mask-aws-account-id: false
```

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,676 @@
...
+          HELPER_IMAGE_REPO: ghcr.io/balena-os/balena-yocto-scripts
+          YOCTO_SCRIPTS_VERSION: ${{ steps.balena-lib.outputs.yocto_scripts_version }}
+          YOCTO_SCRIPTS_REF: ${{ steps.balena-lib.outputs.yocto_scripts_ref }}
+          HELPER_IMAGE_VARIANT: yocto-build-env
+        run: |
+          image_tag="${HELPER_IMAGE_REPO}:${YOCTO_SCRIPTS_VERSION}-${HELPER_IMAGE_VARIANT}"
+          if ! docker pull "${image_tag}"; then
+            image_tag="${HELPER_IMAGE_REPO}:${YOCTO_SCRIPTS_REF}-${HELPER_IMAGE_VARIANT}"
+            docker pull "${image_tag}"
+          fi
```
klutchell, line 630: Try to pull the helper image via the yocto-scripts repo semver.

For open PRs, this semver will not exist on final images yet, only draft images will exist, so fallback to the yocto-scripts git ref. This way we are testing changes to Dockerfiles as part of this PR.

The downside being that the Dockerfiles are untrusted user code and can be modified by forks, which is insecure and relies on human judgement to notice.

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,841 @@
...
+          balenaCloudPassword:
+          #HOSTAPP_PATH: ${{ runner.temp }}/deploy/balena-image.docker
+          VERSION: ${{ steps.balena-lib.outputs.os_version }}
+        run: |
+          set -e
+
+          ## What is that for??
+          # if [ -f "/host/appimage.docker" ]; then
+          #   _local_image=$(docker load -i /host/appimage.docker | cut -d: -f1 --complement | tr -d " " )
+          # fi
```
alexgg, line 596: I don't recall all, but what I remember is that it was made to work for both containers and native, so if there was an `/host/appimage.docker` it means it's running in a container so it uses that image. Probably to do with how hostapp extensions were built at the time.

.github/workflows/yocto-build-deploy.yml
rcooke-warwick: Actually moving the logic into the workflow file instead, its not that bad once simplified (as we know the state we're going to be in once we get to this point)

rcooke-warwick:
Got a first "accidental" full build + deploy to s3 and staging (marked as final by accident)

```
BALENARC_BALENA_URL=balena-staging.com balena os versions generic-amd64
v5.2.4
```

rcooke-warwick:
Note: AMI deploys currently commented out due to not being a priority at the moment

.github/workflows/yocto-build.yml
```diff
@@ -0,0 +1,411 @@
...
+          # Move the generated balena.yml into the deploy folder for easier access in deploy job
+          mv ${WORKSPACE}/balena.yml ${DEPLOY_PATH}
+
+      
+      #TODO:  select correct api key (balena machine or prod)
+
+      - name : Prepare Leviathan inputs
+        run: | 
+          # create "workspace" directory
+          echo LEVIATHAN_WORKSPACE="${WORKSPACE}/leviathan-workspace" >> $GITHUB_ENV
```
rcooke-warwick, line 363: @vipulgupta2048 

This technique doesn't allow using the env var in the same step:
"You can make an environment variable available to any subsequent steps in a workflow job by defining or updating the environment variable and writing this to the GITHUB_ENV environment file. The step that creates or updates the environment variable does not have access to the new value, but all subsequent steps in a job will have access."
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable

.github/workflows/yocto-build-deploy.yml
```diff
@@ -501,7 +501,7 @@ jobs:
           compression-level: 7
           path: |
             ${{ env.DEPLOY_PATH }}/image/balena.img
-            ${{ env.DEPLOY_PATH }}/image/balena-image.docker
+            ${{ env.DEPLOY_PATH }}/balena-image.docker
```
vipulgupta2048, line 504: The hardcoded paths in config.js are very annoying

rcooke-warwick: @vipulgupta2048 we might need to make that configurable for another reason

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,1174 @@
...
+      AUTOKIT_BALENACLOUD_API_KEY:
+        description: balena API key for Leviathan testing from BalenaCloud
+        required: false
+      AUTOKIT_BALENAMACHINE_API_KEY:
+        description: balena API key for Leviathan testing from balena-os BM instance
+        required: false
+      BALENA_API_KEY:
+        description: balena API key for deploying releases # TODO: Different for staging and prod - add another Secret for staging key
+        required: false
+      BALENAOS_STAGING_TOKEN:
```
klutchell, line 17: We will use the GH CLI or API to create the respective secrets on the environments at the repo level rather than have all these secrets for each env.

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,1174 @@
...
+        description: balena API key for deploying releases to staging # TODO: Different for staging and prod - add another Secret for staging key
+        required: false
+      DOCKERHUB_USER:
+        description: Dockerhub user for pulling private helper images
+        required: false
+      DOCKERHUB_TOKEN:
+        description: Dockerhub token for pulling private helper images
+        required: false
```
klutchell, line 25: We can remove these if we stop using the balena-img private helper image for prepare.ts

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,1174 @@
...
+        description: "Comma-separated list of worker types to use for testing. No spaces in between. Example - qemu or testbot."
+        required: false
+        type: string
+        default: "testbot" # Most devices use "testbot" , so qemu is the exception and should be set at the device repo level 
+      BALENACLOUD_APP_NAME:
+        description: "Testbot fleet for finding available Leviathan workers. Not used for QEMU workers. Can accept a list of apps separated by commas, no spaces in between."
+        type: string
+        required: false
+        default: "balena/testbot-rig,balena/testbot-rig-partners,balena/testbot-rig-x86,balena/testbot-rig-partners-x86"
+      BALENACLOUD_ORG: # Why is this an input and not a Var?
```
klutchell, line 143: FIXME

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,1174 @@
...
+  cancel-in-progress: ${{ github.event.action == 'synchronize' }}
+
+env:
+  WORKSPACE: ${{ github.workspace }}
+  MACHINE: ${{ inputs.machine }}
+  VERBOSE: verbose
+  WORKFLOW_NAME: ${{ github.workflow }} # Name of the calling workflow - for use when checking the result of test job on merged PR. Also, can we be clever here and also use it to differentiate between manual/auto runs
+
+permissions:
+  id-token: write   # This is required for requesting the JWT #https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services#requesting-the-access-token
```
klutchell, line 162: Check which permissions we are losing by not specifying the default token permissions here.

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,1174 @@
...
+          done
+
+          echo "::error::External contributions must be approved with the label '${{ env.ok_to_test_label }}'. \
+            Please contact a member of the organization for assistance."
+          exit 1
+
+      # - name: Hotfixing for actuated since they don't have lvm2 installed
+      #   run: |
+      #     sudo apt update
+      #     sudo apt -y install lvm2
```
klutchell, line 239: deleteme

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,1174 @@
...
+          nproc
+          free -h
+          BARYS_ARGUMENTS_VAR="${BARYS_ARGUMENTS_VAR} -a BB_NUMBER_THREADS=${BB_NUMBER_THREADS}"
+          BARYS_ARGUMENTS_VAR="${BARYS_ARGUMENTS_VAR} -a BB_NUMBER_PARSE_THREADS=${BB_NUMBER_PARSE_THREADS}"
+          BARYS_ARGUMENTS_VAR="${BARYS_ARGUMENTS_VAR} -a PARALLEL_MAKE=${PARALLEL_MAKE}"
+          BARYS_ARGUMENTS_VAR="${BARYS_ARGUMENTS_VAR} -a PARALLEL_MAKEINST=${PARALLEL_MAKEINST}"
+          BARYS_ARGUMENTS_VAR="${BARYS_ARGUMENTS_VAR} -a BB_PRESSURE_MAX_CPU=${BB_PRESSURE_MAX_CPU}"
+          BARYS_ARGUMENTS_VAR="${BARYS_ARGUMENTS_VAR} -a BB_PRESSURE_MAX_IO=${BB_PRESSURE_MAX_IO}"
+          BARYS_ARGUMENTS_VAR="${BARYS_ARGUMENTS_VAR} -a BB_PRESSURE_MAX_MEMORY=${BB_PRESSURE_MAX_MEMORY}"
+          echo "BARYS_ARGUMENTS_VAR=${BARYS_ARGUMENTS_VAR}" >> $GITHUB_ENV
```
klutchell, line 410: Delete me when using properly isolated self-hosted runner resources

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,1174 @@
...
+
+      - name: Build
+        id: build
+        env:
+          HELPER_IMAGE_REPO: ghcr.io/balena-os/balena-yocto-scripts
+        run: |
+          # DEBUG - test increasing : vm.mmap_min_addr
+          sysctl vm.mmap_min_addr
+          ulimit
+          sudo sysctl -w vm.mmap_min_addr=65536
```
klutchell, line 453: Can we comment inline why this was required?

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,1174 @@
...
+      # # https://github.com/keithweaver/aws-s3-github-action
+      # - uses: keithweaver/aws-s3-github-action@v1.0.0
+      #   if: inputs.deploy-s3 == true && steps.balena-lib.outputs.deploy_artifact != 'docker-image'
+      #   with:
+      #     command: put
+      #     source: ${{ runner.temp }}/deploy/image/${{ steps.balena-lib.outputs.device_slug }}/latest
+      #     destination: s3://${{ vars.AWS_S3_BUCKET || vars.S3_BUCKET }}/${{ steps.s3-images-dir.outputs.string || steps.s3-esr-images-dir.outputs.string }}/
+      #     aws_access_key_id: ${{ secrets.AWS_SECRET_ACCESS_KEY || secrets.S3_SECRET_KEY}}
+      #     aws_secret_access_key: ${{ secrets.AWS_SECRET_ACCESS_KEY || secrets.S3_ACCESS_KEY }}
+      #     aws_region: ${{ vars.AWS_REGION || vars.S3_REGION }}
```
klutchell, line 618: delete these experimental actions, we are replacing S3 upload with webresources anyway

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,1174 @@
...
+            else
+              $input | split($delim) # Split the input by the delimiter
+            end
+          ')"
+
+          echo "build=${JSON_ARRAY}" >> "${GITHUB_OUTPUT}"
+          echo "json=${JSON_ARRAY}" >> "${GITHUB_OUTPUT}"
+        env:
+          INPUT: ${{ inputs.test-workers }}
+          DELIMITER: ","
```
klutchell, line 989: Can we added a comment with the expected JSON output as an example?

.github/workflows/yocto-build-deploy.yml
rcooke-warwick: This is set to `write` for the sole purpose of getting the AWS token: https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services#requesting-the-access-token

As far as I can tell its a hard requirement for this feature. This is the only thing we lose by not setting it to `write`

.github/workflows/yocto-build-deploy.yml
rcooke-warwick: added as comments in the workflow file

.github/workflows/yocto-build-deploy.yml
rcooke-warwick: added as comments in the workflow file

.github/workflows/yocto-build-deploy.yml
klutchell: Yes, but check this out: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token

> You can use the permissions key in your workflow file to modify permissions for the GITHUB_TOKEN for an entire workflow or for individual jobs. This allows you to configure the minimum required permissions for a workflow or job. When the permissions key is used, all unspecified permissions are set to no access, with the exception of the metadata scope, which always gets read access.

So by setting `id-token: write` and omitting all other permissions, we have removed all other permissions.

You can verify at the top of the `Set up job` step [here](https://github.com/balena-os/balena-yocto-scripts/actions/runs/9322094458/job/25663178018?pr=329).
```
Current runner version: '2.316.1'
Runner name: 'c3ccb3f8-079[2](https://github.com/balena-os/balena-yocto-scripts/actions/runs/9322094458/job/25663178018?pr=329#step:1:2)-4b8b-a4db-cf1cd3b04655'
Runner group name: 'self-hosted'
Machine name: 'f87[3](https://github.com/balena-os/balena-yocto-scripts/actions/runs/9322094458/job/25663178018?pr=329#step:1:3)2db'
GITHUB_TOKEN Permissions
  Metadata: read
Secret source: Actions
Prepare workflow directory
Prepare all required actions
Getting action download info
```

Instead we need [this](https://github.com/product-os/environment-production/blob/890f2e32975f7035511bb9da2fd9938e56f2155e/.github/workflows/flowzone.yml#L9-L25)

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,1174 @@
...
+        required: false
+        type: string
+        default: "us-east-1"
+      test-suites:
+        description: "comma-separated list of test suites to run. No spaces in between."
+        required: false
+        type: string
+        default: "os,hup,cloud"
+      test-workers:
+        description: "Comma-separated list of worker types to use for testing. No spaces in between. Example - qemu or testbot."
```
lmbarros, line 138: The is confusing, as the example does not match the rest of the description.

lmbarros, line 138: I think this is correct:

```suggestion
        description: "Comma-separated list of worker types to use for testing. No spaces in between. The valid worker types are `qemu` and `testbot`."
```

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,1174 @@
...
+          REPORTS: ${{ env.LEVIATHAN_REPORTS }}
+          SUITES: ${{ env.LEVIATHAN_TEST_SUITE }}
+          WORKER_TYPE: ${{ matrix.WORKER_TYPE }}
+          WORKSPACE: ${{ env.LEVIATHAN_WORKSPACE }}
+          TEST_SUITE: ${{ matrix.TEST_SUITE }}
+
+      # Only to run when images are signed, have to run secureboot tests
+      # Make sure it has all the same env variables as the job above
+      - name: "[Secureboot] BalenaOS Leviathan Tests"
+        if: inputs.sign-image == true
```
lmbarros, line 1154: I am confused here. Doesn't the step before this one need `if: inputs.sign-image == false`?

At least to me it seems that either:

* We are wasting resources running the tests twice for secure boot images; or
* The two additional variables we define here trigger some more complex magic on Leviathan, in which case some additional comments could avoid future misunderstandings.

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,1174 @@
...
+      deploy-s3:
+        description: Whether to deploy images to S3
+        required: false
+        type: boolean
+        default: ${{ github.event_name == 'push' }} # Deploy to s3 only on creation of new tags - a "push " event. We're happy with just push events, as the caller workflow only starts this on pushed tags, matched with semver
+      deploy-hostapp:
+        description: Whether to deploy a hostApp container image to a balena environment
+        required: false
+        type: boolean
+        default: true # we always want to deploy the hostapp - as eventually we'll use it for HUP tests
```
lmbarros, line 87: If we always want to deploy them, why is this an option? :-)

Also: Maybe worth rephrasing to express the correct interpretation of "eventually we'll use it for HUP tests". Either of:

1. Some day in the future we'll change the HUP tests to use the hostapp
2. Some step further down the workflow will use the hostapp

.github/workflows/yocto-build-deploy.yml
vipulgupta2048: Yeah and agreed the next suggestion.

.github/workflows/yocto-build-deploy.yml
rcooke-warwick: good catch, thats correct

vipulgupta2048: True that and additional comments are needed which I will add. It is definitely some magic that Leviathan does, good catch

.github/workflows/yocto-build-deploy.yml
rcooke-warwick: Mainly the option was there in case we want to do a manual run of the build and/test - and you'll save a (admittedly small) bit of time. But I see the point!

Agreed on the rewording

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,1174 @@
...
+      # Example output: + json='{"TEST_SUITE":["os","hup","cloud"],"DEVICE_TYPE":["generic-amd64"],"ENVIRONEMENT":["balenaos-balenamachine"],"WORKER_TYPE":["qemu"]}'
+      - name: Create Leviathan test matrix
+        id: leviathan_test_matrix
+        if: inputs.run-tests == true
+        env:
+          MATRIX: >
+            {
+              "TEST_SUITE": ${{ steps.test_suites.outputs.json }},
+              "DEVICE_TYPE": ${{ format('["{0}"]', steps.balena-lib.outputs.device_slug) }},
+              "ENVIRONEMENT": ${{ format('["{0}"]', inputs.test-environment) }},
```
lmbarros, line 1019: This one is also puzzling me a bit. I believe this is a typo, but:

1. The example output above also says `ENVIRONEMENT` like here (possibly just copy and paste).
2. I couldn't spot where this is used to double check it.

```suggestion
              "ENVIRONMENT": ${{ format('["{0}"]', inputs.test-environment) }},
```

.github/workflows/yocto-build-deploy.yml
klutchell: This secret will be deprecated once we deploy the `BALENA_API_KEY` secret to each environment.

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,1174 @@
...
+on:
+  workflow_call:
+    secrets:
+      # API key names should be descriptive telling us what company, what for, where from, what's the key.
+      # Let's say, XCompany wants to have test, then keys could be: XCOMPANY_AUTOKIT_BM_ACCESS_TOKEN
+      # TODO: Use Github CLI or API to fetch the correct key based on environment, instead of having separate secrets for each env
+      AUTOKIT_BALENACLOUD_API_KEY:
+        description: balena API key for Leviathan testing from BalenaCloud
+        required: false
+      AUTOKIT_BALENAMACHINE_API_KEY:
```
klutchell, line 12: This secret will be deprecated once we deploy the `BALENA_API_KEY` secret to each environment.

.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,1174 @@
+name: "Yocto Build-Test-Deploy"
+
+on:
+  workflow_call:
+    secrets:
+      # API key names should be descriptive telling us what company, what for, where from, what's the key.
+      # Let's say, XCompany wants to have test, then keys could be: XCOMPANY_AUTOKIT_BM_ACCESS_TOKEN
+      # TODO: Use Github CLI or API to fetch the correct key based on environment, instead of having separate secrets for each env
+      AUTOKIT_BALENACLOUD_API_KEY:
```
klutchell, line 9: This secret will be deprecated once we deploy the `BALENA_API_KEY` secret to each environment.

vipulgupta2048: SGTM

.github/workflows/yocto-build-deploy.yml
klutchell: fixed!

vipulgupta2048: Hi, thanks that's my english clearly failing me sometimes. I misspell environment all the time with that extra E. 
It happen sometimes when I am writing `levaithan` too much instead of `leviathan`

.github/workflows/yocto-build-deploy.yml
rcooke-warwick: Added `statuses: read` as I think thats the only extra permission we need for `GITHUB_TOKEN` - to read the status checks of the merge commits to see if tests passed. 

We can add other permissions as we discover we need them

rcooke-warwick:
one last note, I had to make the device repo checkout in the leviathan step recursive to ensure we actually fetched the leviathan submodule cc @vipulgupta2048  , as its buried 2 modules deep - we didn't see that error before because we were specifying a meta-balena ref in the `generic-amd64.yml` workflow file, and getting leviathan from that part.

klutchell:
I self-certify!

PR #336, opened 2024-04-19
balena-deploy: deploy usbboot if available

alexgg:
This is where the RPI family deploys provisioning artifacts.

Change-type: patch


alexgg:
Tested a local update from v2.51 to this PR version.

klutchell (approved):

PR #337, opened 2024-04-22
Add Leviathan test workflows

vipulgupta2048:


.github/workflows/yocto-build.yml
```diff
@@ -3,11 +3,19 @@ name: "Yocto"
...
   workflow_call:
     secrets:
+      # API key names should be descriptive telling us what company, what for, where from, what's the key.
+      # Let's say, XCompany wants to have test, then keys could be: XCOMPANY_AUTOKIT_BM_ACCESS_TOKEN
+      AUTOKIT_BALENACLOUD_API_KEY:
+        description: balena API key for Leviathan testing from BalenaCloud
+        required: false
+      AUTOKIT_BALENAMACHINE_API_KEY:
+        description: balena API key for Leviathan testing from balena-os BM instance
+        required: false
```
klutchell, line 13: Do we need these to be separate? Why not use a single secret and have ops change it if/when we move autokit back to balenaCloud.

What about `AUTOKIT_FLEET_API_KEY` so we know what it's used for but is environment agnostic?

.github/workflows/yocto-build.yml
```diff
@@ -63,13 +71,27 @@ on:
...
-    outputs:  
-      device_slug: 
+      QEMU_CPUS:
+        required: false
+        type: number
+        default: 4
+      QEMU_MEMORY:
+        type: string
+        required: false
+        default: "1G"
```
klutchell, line 81: Let's hardcode these values in the workflow where we call the action. Only 2-3 repositories use QEMU for testing anyway so these fields are useless most of the time.

.github/workflows/yocto-build.yml
```diff
@@ -408,4 +401,179 @@ jobs:
...
+        - DEVICE_TYPE: ${{ needs.build.outputs.device_slug }}
+          WORKER_TYPE: qemu
+          ENVIRONMENT: ${{ inputs.environment }}
+          TEST_SUITE: hup
+        - DEVICE_TYPE: ${{ needs.build.outputs.device_slug }}
+          WORKER_TYPE: qemu
+          ENVIRONMENT: ${{ inputs.environment }}
+          TEST_SUITE: cloud
+    steps:
+      - name: Reject unapproved external contributions
```
klutchell, line 437: Since this job depends on the `build` job, we really only need to add this rejection step to the build.

.github/workflows/yocto-build.yml
```diff
@@ -408,4 +401,179 @@ jobs:
...
+        uses: easimon/maximize-build-space@fc881a613ad2a34aca9c9624518214ebc21dfc0c
+        with:
+          root-reserve-mb: "4096"
+          temp-reserve-mb: "1024"
+          swap-size-mb: "4096"
+          remove-dotnet: "true"
+          remove-android: "true"
+          remove-haskell: "true"
+          remove-codeql: "true"
+          remove-docker-images: "true"
```
klutchell, line 478: Shouldn't need this for testing, yocto just happens to need a lot of space

.github/workflows/yocto-build.yml
```diff
@@ -408,4 +401,179 @@ jobs:
...
+          git submodule update --init --recursive
+
+      - name: Setup env variables for Prod
+        if: inputs.environment == 'production'
+        run: |
+          echo "BALENACLOUD_API_KEY=${{ secrets.AUTOKIT_BALENACLOUD_API_KEY  }}" >> $GITHUB_ENV
+          echo "BALENACLOUD_API_URL='balena-cloud.com'" >> $GITHUB_ENV
+          echo "BALENACLOUD_ORG='testbot'" >> $GITHUB_ENV
+          echo "BALENACLOUD_APP_NAME='balena/testbot-rig'" >> $GITHUB_ENV
+          echo "BALENACLOUD_SSH_PORT=???" >> $GITHUB_ENV
```
rcooke-warwick, line 514: there is a built in leviathan default for prod balena cloud 

otherwise, the balenamachine  github environment should have a `BALENACLOUD_SSH_PORT` , the production one should have a `BALENACLOUD_SSH_PORT` , and in the workflow here we just reference {{var.BALENACLOUD_SSH_PORT}}

.github/workflows/yocto-build.yml
```diff
@@ -408,4 +401,179 @@ jobs:
...
+      # If are not able to tweak, HUP suite to use hostapp instead then 
+      # We need to do a per matrix addition to make sure we, 
+      # Pull balena.img.gz for os, cloud
+      # Pull balena-image.docs for HUP with exactly these names
+      ## Ryan
+      - name: Fetch artifacts from build job
+        uses: actions/download-artifact@v4
+        with:
+          name: build-artifacts # Ryan will tell what the name of the artifact would become
+          path: ${{ env.LEVIATHAN_WORKSPACE }}
```
rcooke-warwick, line 559: in this case I beleive it will become ${LEVIATHAN_WORKSPACE}/balena.img

.github/workflows/yocto-build.yml
rcooke-warwick: The issue at the moment @klutchell is that we have a split fleet, some autokits are balena cloud, some balena machine. I agree these shouldn't be separate though

I think what we actually want here is a step further down, that selects the correct API key for the environment

klutchell: That's unfortunate. What we should do then, is add `environment` to the Leviathan test matrix so when we build the matrix we will populate a value for environment then the appropriate vars can be read in the test run.

This doesn't solve secrets unfortunately unless we add them manually to all the environments, or GitHub starts supporting org-level environments.

PR #340, opened 2024-06-04
[TEST ONLY] Playing around with GHA

lmbarros:


.github/workflows/yocto-build-deploy.yml
```diff
@@ -0,0 +1,1126 @@
...
+      # Example output: + json='{"TEST_SUITE":["os","hup","cloud"],"DEVICE_TYPE":["generic-amd64"],"ENVIRONEMENT":["balenaos-balenamachine"],"WORKER_TYPE":["qemu"]}'
+      - name: Create Leviathan test matrix
+        id: leviathan_test_matrix
+        if: inputs.run-tests == true
+        env:
+          MATRIX: >
+            {
+              "TEST_SUITE": ${{ inputs.test-suites }},
+              "DEVICE_TYPE": ${{ format('["{0}"]', steps.balena-lib.outputs.device_slug) }},
+              "ENVIRONEMENT": ${{ format('["{0}"]', inputs.test-environment) }},
```
lmbarros, line 971: This one is also puzzling me. I believe this is a typo, but:

1. The example output above also says `ENVIRONEMENT` like here (possibly just copy and paste).
2. I couldn't spot where this is used to double check it.

```suggestion
              "ENVIRONMENT": ${{ format('["{0}"]', inputs.test-environment) }},
```

PR #341, opened 2024-06-06
Add catchall job to yocto-build-deploy for merge requirements

klutchell:
This is a helper job to avoid having to mark all build/test matrices as required, as this job will always run and return success or failure.

Change-type: patch

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -1192,3 +1192,40 @@ jobs:
           # Two variables extra for this job
           QEMU_SECUREBOOT: 1
           FLASHER_SECUREBOOT: 1
+
+  # This job always runs and will fail if any of the builds or tests fail.
+  # This way we can mark this job as required for merging PRs.
+  # Otherwise we would need to mark each build and test matrix, suite, etc. as required.
+  all_jobs:
+    name: All jobs
+    needs:
+      - build
+      - test
+    runs-on: ubuntu-latest
+    # Do not skip this job if needed jobs are skipped, as we always
+    # want to capture pass/fail for the entire workflow and skipped
+    # jobs count as passing as far as required checks go.
+    if: |
+      always() && inputs.run-tests == true
+
+    defaults:
+      run:
+        working-directory: .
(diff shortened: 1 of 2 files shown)


.github/workflows/yocto-build-deploy.yml
```diff
@@ -1192,3 +1192,40 @@ jobs:
...
+    name: All jobs
+    needs:
+      - build
+      - test
+    runs-on: ubuntu-latest
+    # Do not skip this job if needed jobs are skipped, as we always
+    # want to capture pass/fail for the entire workflow and skipped
+    # jobs count as passing as far as required checks go.
+    if: |
+      always() && inputs.run-tests == true
```
lmbarros, line 1209: Isn't `always()` redundant here?

klutchell: We need `always()` because the hidden defaults are `success()`, and we want to make sure this job always runs no matter the results of previous jobs.

If this job gets skipped because a previous job failed, GH would actually consider `All tests` to be passed and will merge the PR. I think they have their reasons for doing it that way but I don't recall what they are.

klutchell: I do have to remove `inputs.run-tests == true` though, that was before I included the build in this check.

vipulgupta2048 (commented):
LGTM

.github/workflows/yocto-build-deploy.yml
```diff
@@ -1192,3 +1192,40 @@ jobs:
...
+      run:
+        working-directory: .
+        shell: bash --noprofile --norc -eo pipefail -x {0}
+
+    steps:
+      - name: Reject failed jobs
+        run: |
+          if [ "${{ contains(needs.*.result, 'failure') }}" = "true" ]
+          then
+            echo "One or more tests have failed"
```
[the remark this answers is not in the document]
vipulgupta2048: This seems like a very good thing Flowzone does too, any other checks or bounds we can bring in from that project.

.github/workflows/yocto-build-deploy.yml
klutchell: I've updated the comment!

PR #347, opened 2024-06-07
Simplify matrix generation

lmbarros:
This is a more polished (and working!) version of my previous attempt to simplify the matrix generation.

Turns out my change by itself was OK, but I missed that `generic-amd64.yml` was setting `inputs.test-workers` to a value that was not in the format now expected. Not the commit updates this input, too.

And then I included a second commit that validates the inputs early on the workflow.

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -132,15 +132,24 @@ on:
         type: string
         default: "us-east-1"
       test-suites:
-        description: "Comma-separated list of test suites to run. No spaces in between."
+        description: "JSON list with the test suites to run."
         required: false
         type: string
-        default: "os,hup,cloud"
+        default: >
+          [
+            "os",
+            "hup",
+            "cloud"
+          ]
       test-workers:
-        description: "Comma-separated list of worker types to use for testing. No spaces in between. The valid worker types are `qemu` and `testbot`."
+        description: "JSON list of worker types to use for testing. Valid element values are `qemu` and `testbot`."
         required: false
         type: string
-        default: "testbot" # Most devices use "testbot" , so qemu is the exception and should be set at the device repo level
+        # Most devices use "testbot", so "qemu" is the exception and should be set at the device repo level
+        default: >
+          [
+            "testbot"
+          ]
       worker-fleets:
         description: "Testbot fleet for finding available Leviathan workers. Not used for QEMU workers. Can accept a list of apps separated by commas, no spaces in between."
         type: string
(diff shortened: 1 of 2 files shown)


.github/workflows/yocto-build-deploy.yml
```diff
@@ -132,15 +132,24 @@ on:
...
-        description: "Comma-separated list of worker types to use for testing. No spaces in between. The valid worker types are `qemu` and `testbot`."
+        description: "JSON list of worker types to use for testing. Valid element values are `qemu` and `testbot`."
         required: false
         type: string
-        default: "testbot" # Most devices use "testbot" , so qemu is the exception and should be set at the device repo level
+        # Most devices use "testbot", so "qemu" is the exception and should be set at the device repo level
+        default: >
+          [
+            "testbot"
+          ]
```
klutchell, line 152: @vipulgupta2048 is the idea here that if we provide more than one we will get an additional dimension in the matrix and run 6 test suites?

example
```json
{"TEST_SUITE":["os","hup","cloud"],"DEVICE_TYPE":["generic-amd64"],"ENVIRONMENT":["balenaos-balenamachine"],"WORKER_TYPE":["testbot","qemu"]}
```
results in 6 jobs
- `generic-amd64,balenaos-balenamachine,os,testbot`
- `generic-amd64,balenaos-balenamachine,os,qemu`
- `generic-amd64,balenaos-balenamachine,hup,testbot`
- `generic-amd64,balenaos-balenamachine,hup,qemu`
- `generic-amd64,balenaos-balenamachine,cloud,testbot`
- `generic-amd64,balenaos-balenamachine,cloud,qemu`

Similarly, adding another environment would result in 12 jobs. Is this the kind of matrix we were aiming for?

I wonder if we couldn't simplify @lmbarros PR further by having a test matrix as a JSON input which expects an `includes` array.

example default input:
```json
[
  {
    "TEST_SUITE": "os",
    "ENVIRONMENT": "balenaos-balenamachine",
    "WORKER_TYPE": "testbot"
  },
  {
    "TEST_SUITE": "cloud",
    "ENVIRONMENT": "balenaos-balenamachine",
    "WORKER_TYPE": "testbot"
  },
  {
    "TEST_SUITE": "hup",
    "ENVIRONMENT": "balenaos-balenamachine",
    "WORKER_TYPE": "testbot"
  },
]
```

The DEVICE_TYPE is already available via the build and other inputs/outputs on the workflow and should not be required as part of the input matrix as it will never change from the build, or be more than one device.

Then that entire, unmodified, JSON blob can be used in the matrix like this:
```yml
  strategy:
      fail-fast: false
      matrix:
        includes: ${{ fromJSON(inputs.test_matrix) }}
```
IMO this provides the most control and is very flexible for the calling workflow, while keeping the results predicable. Each object in the array results in a test run, but the suite, environment, and worker-type are configurable at a per-run level.

klutchell (changes requested):

ab77:
@lmbarros `BALENAOS_STAGING_TOKEN` is mutually exclusive with the "environment" approach. When this workflow runs in `balena-staging.com` GitHub env., `(BALENA)_API_KEY` secret will be set to the staging value. There is no longer a need to keep this distinction in the workflow.

ab77 (changes requested):
See https://github.com/balena-os/balena-yocto-scripts/pull/347#issuecomment-2160920186

lmbarros:
@ab77 , I don't think this point you mentioned here is related to this PR, is it?

ab77:
> @ab77 , I don't think this point you mentioned here is related to this PR, is it?

I am pointing out it will possibly break this PR if not addressed.

klutchell:
Superseded by https://github.com/balena-os/balena-yocto-scripts/pull/350

PR #349, opened 2024-06-12
Update the triggers on the example test workflow

klutchell:
We do not need to run the workflow on close/merge, and we can skip the manual workflow runs on this project as each device type will have it's own workflow in the device repos supporting manual triggers.

Change-type: patch


lmbarros (approved):
:+1:

PR #352, opened 2024-06-17
Refactor secrets and variables to use environments

klutchell:
Reduce the required secrets and inputs required by relying on vars and secrets set in each GitHub Environment.

Change-type: patch

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -3,20 +3,15 @@ name: "Yocto Build-Test-Deploy"
 on:
   workflow_call:
     secrets:
-      # TODO: Deprecate this secret in favor of the BALENA_API_KEY secret set at the environment level
-      AUTOKIT_BALENACLOUD_API_KEY:
-        description: balena API key for Leviathan testing from BalenaCloud
+      # BALENA_API_DEPLOY_KEY is a secret that should be specific to the runtime environment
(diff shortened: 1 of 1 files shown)


.github/workflows/yocto-build-deploy.yml
```diff
@@ -983,9 +957,9 @@ jobs:
 
     env:
       # Variables provided via the selected GitHub environment
-      BALENACLOUD_API_URL: ${{ vars.BALENA_URL || matrix.environment }}
+      BALENACLOUD_API_URL: ${{ matrix.environment }}
       BALENACLOUD_SSH_PORT: ${{ vars.BALENACLOUD_SSH_PORT || '222' }}
-      BALENACLOUD_SSH_URL: ssh.devices.${{ vars.BALENA_URL || matrix.environment }}
+      BALENACLOUD_SSH_URL: ssh.devices.${{ matrix.environment }}
```
vipulgupta2048, line 962: So, just to be sure, we are not going with creating a balenaMachine environment in GitHub now for only testing to help us ease up on the mass configuration and instead people can explicitly provide an environment now in the `test_matrix`

Correct?

klutchell: So `matrix.environment == URL == GitHub Environment name`

vipulgupta2048: Sounds good.

vipulgupta2048 (approved):

PR #355, opened 2024-06-27
GHA workflow: Simplify check for secure boot

lmbarros:
I spotted a typo/bug in the check for secure boot this while [checking this test run](https://github.com/balena-io-experimental/balena-raspberrypi-gha/actions/runs/9681366732/job/26756255483#step:29:27):

```
# Mixing a check for string length with an equality check!
if [ -n "${{ inputs.sign-image }}" = "true" ]; then
```

Then I noticed we could use `inputs.sign-image` directly, without going through a new variable `SECURE_BOOT_FEATURE_FLAG`. This also removes the small complexity of one thing being true/false and the other being yes/no.

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -602,15 +602,6 @@ jobs:
       # hostapp Deploy
       ##############################
 
-      - name: Set SECURE_BOOT_FEATURE_FLAG
-        if: inputs.deploy-hostapp == true
-        run: |
-          if [ -n "${{ inputs.sign-image }}" = "true" ]; then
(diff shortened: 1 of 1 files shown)


klutchell:
Good catch! I had to go and check that `SECURE_BOOT_FEATURE_FLAG` and it's not, as we copied the logic from the deploy scripts into the workflow verbatim.

That's how we ended up with this redundancy, a previous iteration of the workflow was calling the deploy bash scripts.

vipulgupta2048 (commented):
LGTM, SECURE_BOOT_FEATURE_FLAG is not being used in the caller workflow directly or anywhere else in the workflow? Good improvement.

vipulgupta2048 (approved):

PR #356, opened 2024-07-01
Fix quoting of $GITHUB_OUTPUT

lmbarros:
We had typos in two cases, in which we missed the opening quote.

This seems to be final blocker preventing a complete execution of the workflow!


klutchell (approved):

ab77 (approved):
This is another use case for enabling shellcheck for actionslint @klutchell

PR #357, opened 2024-07-02
Stop uploading to GH artifacts storage when PR is closed

vipulgupta2048:
Uploading artifacts to GH artifact storage when PR is merged (closed event) is wasterful. Since they are only uploaded for temporary basis so it can be used for testing. Hence, the PR to stop doing that for `closed` PR event

Signed-off-by: Vipul Gupta (@vipulgupta2048) <vipulgupta2048@gmail.com>

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -503,6 +503,7 @@ jobs:
       # Separate "flasher" and "raw" variants are not used in the testing flow
       - name: Upload artifacts
         uses: actions/upload-artifact@26f96dfa697d77e81fd5907df203aa23a56210a8 # v4.3.0
+        if: github.event_name != 'push'
         with:
           name: build-artifacts
           if-no-files-found: error


.github/workflows/yocto-build-deploy.yml
```diff
@@ -503,6 +503,7 @@ jobs:
       # Separate "flasher" and "raw" variants are not used in the testing flow
       - name: Upload artifacts
         uses: actions/upload-artifact@26f96dfa697d77e81fd5907df203aa23a56210a8 # v4.3.0
+        if: github.event.pull_request.state !== 'closed'
```
klutchell, line 506:
```suggestion
        if: github.event.pull_request.state != 'closed'
```

vipulgupta2048: Javascipt brain for strict equality. My bad.

.github/workflows/yocto-build-deploy.yml
```diff
@@ -503,6 +503,7 @@ jobs:
       # Separate "flasher" and "raw" variants are not used in the testing flow
       - name: Upload artifacts
         uses: actions/upload-artifact@26f96dfa697d77e81fd5907df203aa23a56210a8 # v4.3.0
+        if: github.event.pull_request.state != 'closed'
```
klutchell, line 506:
```suggestion
        if: github.event_name != 'push'
```
This workflow doesn't listen for PR close events, so your existing IF condition would be no-op and we would still upload on tag events, which we don't want IIRC.

klutchell (changes requested):

klutchell (approved):

.github/workflows/yocto-build-deploy.yml
vipulgupta2048: I see what you mean, the called workflow doesn't listen to the `pull_request` closed event since that's for the caller workflow isn't it.

PR #358, opened 2024-07-02
Remove dry-run flag from S3 upload

klutchell:
We are publishing hostapp releases to staging already, we should start including the associated S3 files as well.

Change-type: patch


vipulgupta2048 (commented):
Would it be better to keep this around and only merge when we are actually merging balena-intel for instance which I understand should be soon soon.

rcooke-warwick (approved):

klutchell:
> Would it be better to keep this around and only merge when we are actually merging balena-intel for instance which I understand should be soon soon.

We are already deploying (and sometimes finalizing) hostapp releases to staging, so I would prefer we keep them aligned with the S3 images. Otherwise we could break e2e tests with generic-amd64 for example.

vipulgupta2048 (approved):

PR #359, opened 2024-07-03
use workflow runs of PR to get test results

rcooke-warwick:
Since we have moved to workflows for tests instead of status checks, we have to fetch the workflow runs for the appropriate commit to determine a test pass/fail.

notable points for review:
1. as we're using workflow names to find the appropriate run, worflow name in the device repo must be named after machine name - updated the example here to reflect that
2. If the test matrix is empty, don't run the `check-merge-tests` step - and therefore do not finalise the hostapp. Is this a valid way of doing that. This is to stop device's releases being finalized without undergoing tests

Change-type: patch

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -84,11 +84,6 @@ on:
         required: false
         type: boolean
         default: false # The default is "no" - because the `check-merge-tests` will determine whether or not to mark the hostapp as final. This is purely here for a manual override
-      check-merge-tests:
-        description: Whether to check the test results from the merge commit that resulted in new tagged version - can be overridden in dispatch for manual deploy
-        required: false
-        type: boolean
-        default: ${{ github.event_name == 'push' }} # This determines if we want to check the results of the merge PR - we only want to do it when a new tag is made
       deploy-ami:
         description: Whether to deploy an AMI to AWS
         required: false
@@ -156,14 +151,15 @@ env:
   WORKSPACE: ${{ github.workspace }}
   MACHINE: ${{ inputs.machine }}
   VERBOSE: verbose
-  WORKFLOW_NAME: ${{ github.workflow }} # Name of the calling workflow - for use when checking the result of test job on merged PR. Also, can we be clever here and also use it to differentiate between manual/auto runs
+  WORKFLOW_NAME: ${{ github.workflow }}
 
 # https://docs.github.com/en/actions/security-guides/automatic-token-authentication
 # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions
 # https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#adding-permissions-settings
 permissions:
-  id-token: write # This is required for requesting the JWT #https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services#requesting-the-access-token
-  statuses: read # We are fetching status check results of a merge commit when workflow is triggered by new tag, to see if tests pass
+  id-token: write   # This is required for requesting the JWT #https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services#requesting-the-access-token
+  actions: read    # We are fetching workflow run results of a merge commit when workflow is triggered by new tag, to see if tests pass
+  pull-requests: read # Required to fetch the PR that merged, in order to get the test results
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/yocto-build-deploy.yml
```diff
@@ -293,19 +294,30 @@ jobs:
       #     regex: '^v20[0-9][0-9].[0-1]?[1470].[0-9]+$'
 
       # This will control the deployment of the hostapp only - it will determine if it is marked as final or not
-      # The hostapp being finalised is what determines if the API will present this OS version to users
+      # The hostapp being finalised is what determines if the API will present this OS version to user
+      # If the test_matrix is empty - it means there are no tests for the DT - so don't check tests, and don't finalise
+      # This check will check for a workflow with the same name as the yocto machine - so ensure that workflows are named after the yocto-machine
       - name: Check test results
-        if: ${{ inputs.check-merge-tests }} # Left in the case of manual deploys where tests are failing but we had to force merge
+        if: ${{ inputs.check-merge-tests && inputs.test_matrix != '{}' }} # Left in the case of manual deploys where tests are failing but we had to force merge
```
vipulgupta2048, line 301: The input test matrix won't be empty; we have a default set for it.
So, on merge, it will have the default value. 

Or before this PR, we merge the poll option that we all agreed on `Default to no tests unless provided via `test_matrix` JSON and deprecate `run_tests` boolean`

.github/workflows/yocto-build-deploy.yml
```diff
@@ -293,19 +294,30 @@ jobs:
...
         run: |
+          # Gets the PR number of the merge commit
           prid=$(gh api -H "Accept: application/vnd.github+json" /repos/$REPO/commits/$COMMIT --jq '.commit.message' | head -n1 | cut -d "#" -f2 | awk '{ print $1}')
-          status_url=$(gh api -H "Accept: application/vnd.github+json" /repos/$REPO/pulls/$prid --jq '._links.statuses.href')
+          
+          # Gets the head commit of the PR - needed to fetch workflows ran on that commit
+          head=$(gh api -H "Accept: application/vnd.github+json" /repos/$REPO/pulls/$prid --jq '.head.sha')
+
+          # Fetching workflow runs and filtering by the commit of the head of the PR returns the latest attempts of the workflow for that commit
+          # Selecting only for pull request events is because there are skipped pull_request_target triggered checks sometimes (WHY?)
```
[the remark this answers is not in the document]
rcooke-warwick: this was to open up the question to the reviewers + internally think.

.github/workflows/generic-amd64.yml
```diff
@@ -5,7 +5,8 @@
 # to the inputs and the defaults.
 
 # TODO: We need a system to keep these inputs aligned across all device repos
-name: Generic x86_64 (GPT)
+# Name workflow after machine
```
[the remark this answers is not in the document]
klutchell: Use `${{ github.workflow }}` in the search rather than depending on a specific syntax for names. What if we need to build for the same machine in different modes?

.github/workflows/yocto-build-deploy.yml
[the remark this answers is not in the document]
klutchell: The answer to that comment is essentially [here](https://github.com/balena-os/balena-yocto-scripts/blob/da6fe23e89509549866b16cbc6ff404980a189c0/.github/workflows/generic-amd64.yml#L35-L38).

We can repeat ourselves and add an updated comment in the code if it helps.

> We trigger on both PR and PRT events, as the former allows modification of workflows, and the latter allows forks to read GitHub secrets. However we never want to run duplicate workflows, so we add an if condition to make sure the PRT is skipped for internal pull requests, and PR is skipped for external pull requests. This results in there always being both a skipped version of the workflow, and an executed version.

.github/workflows/yocto-build-deploy.yml
klutchell, line 301:
```suggestion
        # https://docs.github.com/en/actions/learn-github-actions/expressions#functions
        # this expression checks if test matrix is any variation of empty
        if: github.event_name == 'push' && contains(fromJSON('["[]","{}","","null"]'), inputs.test_matrix) == false
```

.github/workflows/yocto-build-deploy.yml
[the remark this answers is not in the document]
klutchell: So we need to check for both PR and PRT in the API call, and only one will have All Jobs == success

.github/workflows/yocto-build-deploy.yml
```diff
@@ -293,19 +294,30 @@ jobs:
...
-      # The hostapp being finalised is what determines if the API will present this OS version to users
+      # The hostapp being finalised is what determines if the API will present this OS version to user
+      # If the test_matrix is empty - it means there are no tests for the DT - so don't check tests, and don't finalise
+      # This check will check for a workflow with the same name as the yocto machine - so ensure that workflows are named after the yocto-machine
       - name: Check test results
-        if: ${{ inputs.check-merge-tests }} # Left in the case of manual deploys where tests are failing but we had to force merge
+        if: ${{ inputs.check-merge-tests && inputs.test_matrix != '{}' }} # Left in the case of manual deploys where tests are failing but we had to force merge
         id: merge-test-result
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
klutchell, line 304:
```suggestion
          # environment variables used by gh CLI
          # https://cli.github.com/manual/gh_help_environment
          GH_DEBUG: "true"
          GH_PAGER: "cat"
          GH_PROMPT_DISABLED: "true"
          GH_REPO: "${{ github.repository }}"
          GH_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
```

.github/workflows/yocto-build-deploy.yml
```diff
@@ -257,44 +253,41 @@ jobs:
...
-      # - name: 'Fetch latest tag'
-      #   id: get-latest-tag
-      #   if: ${{ inputs.check-merge-tests }}
-      #   uses: "actions-ecosystem/action-get-latest-tag@v1"
-
       # We're also checking out the tag in this step, so the subsequent build is done from the tagged version of the device repo
       - name: "Fetch merge commit"
         id: set-merge-commit
-        if: ${{ inputs.check-merge-tests }} # Left in the case of manual deploys where tests are failing but we had to force merge
+        if: ${{ github.event_name == 'push' }} # Left in the case of manual deploys where tests are failing but we had to force merge
```
klutchell, line 259: This comment needs updating, essentially we only want to check for previous runs on a tag event, otherwise we will use the boolean value for finalize, which is false by default.

.github/workflows/generic-amd64.yml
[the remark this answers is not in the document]
klutchell: We can remove this change to the workflow name now

.github/workflows/yocto-build-deploy.yml
klutchell: These variables provide a cleaner experience using GH CLI in non-interactive scripts.

klutchell:
The first commit can be dropped and rebased on master now that the actionlint stuff has merged.

.github/workflows/yocto-build-deploy.yml
```diff
@@ -949,6 +948,7 @@ jobs:
     # we should consider using GitHub hosted runners for the testbot workers.
     runs-on: ${{ matrix.runs_on || fromJSON('["self-hosted", "X64", "kvm"]') }}
     environment: ${{ matrix.environment }}
+    if: ${{ github.event_name == 'pull_request' ||  github.event_name == 'pull_request_target'}}
```
klutchell, line 951:
```suggestion
    if: github.event_name != 'push'
```
We need to allow tests to run on workflow dispatch. The only time we don't want tests is on push.

klutchell (changes requested):

klutchell (approved):

PR #361, opened 2024-07-03
Fix actionlint errors and warnings in shell steps

klutchell:
Change-type: patch

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -214,7 +214,7 @@ jobs:
 
           pr_labels="$(gh pr view ${{ github.event.pull_request.number }} --json labels -q .labels[].name)"
 
-          for label in "${pr_labels}"
+          for label in ${pr_labels}
           do
             if [[ "$label" =~ "${{ env.ok_to_test_label }}" ]]
             then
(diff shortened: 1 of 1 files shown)


ab77 (approved):

.github/workflows/yocto-build-deploy.yml
```diff
@@ -696,7 +696,7 @@ jobs:
             # https://github.com/balena-os/balena-yocto-scripts/blob/master/automation/include/balena-api.inc#L914
             # Give the team developer access to the app
             _json=$(${CURL} -XPOST "https://api.${API_ENV}/${TRANSLATION}/team_application_access" -H "Content-Type: application/json" -H "Authorization: Bearer ${BALENAOS_TOKEN}" --data '{"team": "${_team_id}", "grants_access_to__application": "${_appID}", "application_membership_role": "${_role_id}""}')
-            id=$(echo "${_json}" | jq -r '.id')
+            _id=$(echo "${_json}" | jq -r '.id')
```
lmbarros, line 699: Can't we just remove this line? Doesn't look like `id` (or `_id`) is used anywhere.

klutchell: I noticed that too, we need to check and see why it was copied from the original script. We can probably delete it but I didn't dig very far.

PR #364, opened 2024-07-12
Fix os contract building for private device types

rcooke-warwick:
when building + deploying or a private DT , the check to see if the DT is private fails. This is due to https://github.com/balena-os/balena-yocto-scripts/blob/master/automation/include/balena-api.inc#L424 using this function: https://github.com/balena-os/balena-yocto-scripts/blob/master/automation/include/balena-lib.inc#L191 - which uses the jenkins deployTo variable to select the correct api url and token.

Change-type: patch

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -362,6 +362,10 @@ jobs:
       # A lot of outputs inferred from here are used everywhere else in the workflow
       - name: Set build outputs
         id: balena-lib
+        env: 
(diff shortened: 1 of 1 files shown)


klutchell (approved):

rcooke-warwick:
set to draft to prevent merge - testing with :   looks like there's a similar issue with `balena_lib_build_contract` which is used further down in the workflow

klutchell:
@rcooke-warwick can you rebase this so we can get it merged?

vipulgupta2048 (approved):

PR #365, opened 2024-07-12
Use App Installation tokens so we can clone private submodules

klutchell:
Change-type: patch

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -29,6 +29,9 @@ on:
       SIGN_API_KEY:
         description: balena API key that provides access to the signing server
         required: false
+      GH_APP_PRIVATE_KEY:
+        description: "GPG Private Key for GitHub App to generate ephemeral tokens (used with vars.FLOWZONE_APP_ID)"
+        required: false
     inputs:
       build-runs-on:
         description: The runner labels to use for the build job(s)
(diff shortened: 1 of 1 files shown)


klutchell:
@rcooke-warwick this is ready for review!

.github/workflows/yocto-build-deploy.yml
```diff
@@ -238,12 +241,35 @@ jobs:
...
+        id: app-token
+        if: vars.FLOWZONE_APP_ID != ''
+        with:
+          app-id: ${{ vars.FLOWZONE_APP_ID }}
+          private-key: ${{ secrets.GH_APP_PRIVATE_KEY }}
+
+      # Generate another app token for the balena-io organization
+      # so we can checkout private contracts
+      # https://github.com/actions/create-github-app-token
+      - name: Create GitHub App installation token (balena-io)
```
rcooke-warwick, line 258: Are there any security concerns with having a balena-io GH token for all or these jobs, rather than just those that need it? (at the moment that's only private DTs , in order to clone the private-contracts repo , and the leviathan job again to clone private contracts?)

We have to do it as it stands, the question is just is it a problem that we're doing it for all jobs, not just those that require it? Maybe a nitpick but thought I'd ask just in case

klutchell: Good question, and the short answer is this PR doesn't change our security posture in that regard. Those GitHub App secrets are available across all orgs for Flowzone to use, so whether we make use of them or not in this workflow doesn't change much.

One thing we can do is add `persist-credentials: false` so the credentials do not live in the workspace beyond the checkout steps. I've made the change.

rcooke-warwick (approved):

PR #366, opened 2024-07-15
Change finalize-hostapp to tri-state

klutchell:
Value can be truthy, or !truthy, or 'auto' to
check for passing tests on tag events.

This allows us to fully disable the auto-finalize for
some customers that don't want to finalize without
manual testing.

Change-type: minor
See: https://balena.zulipchat.com/#narrow/stream/345889-balena-io.2Fos/topic/2024.2E7.20ESR/near/451498822

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -82,11 +82,14 @@ on:
         # We want to deploy the hostapp by default - as a draft if its on a PR, or as final if a new tagged version. This is an input however to allow for manual runs where deploying the hostapp isn't wanted or needed.
         # At some point in the future we want to modify the HUP test suite to use the hostapp as the HUP target, rather than sending the DUT the docker image and doing a HUP using that file
         default: true
+      # The default is "auto" and `check-merge-tests` will determine whether or not to mark the hostapp as final.
+      # True values include case-insensitive true|yes|1|y
+      # False values include case-insensitive false|no|0|n
       finalize-hostapp:
-        description: Whether to finalize a hostApp container image to a balena environment
+        description: Whether to finalize the hostApp (either 'yes' or 'no' or 'auto' to check for tests on tag events)
         required: false
-        type: boolean
-        default: false # The default is "no" - because the `check-merge-tests` will determine whether or not to mark the hostapp as final. This is purely here for a manual override
+        type: string
+        default: auto
       deploy-ami:
         description: Whether to deploy an AMI to AWS
         required: false
@@ -297,7 +300,10 @@ jobs:
       - name: Check test results
         # https://docs.github.com/en/actions/learn-github-actions/expressions#functions
         # this expression checks if test matrix is any variation of empty
-        if: ${{github.event_name == 'push' && contains(fromJSON('["[]","{}","","null"]'), inputs.test_matrix) == false }}
+        if: |
+          inputs.finalize-hostapp == 'auto' &&
+          github.event_name == 'push' && 
+          contains(fromJSON('["[]","{}","","null"]'), inputs.test_matrix) == false
         id: merge-test-result
         env:
           REPO: ${{ inputs.device-repo }}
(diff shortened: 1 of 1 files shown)


.github/workflows/yocto-build-deploy.yml
```diff
@@ -82,11 +82,14 @@ on:
         # We want to deploy the hostapp by default - as a draft if its on a PR, or as final if a new tagged version. This is an input however to allow for manual runs where deploying the hostapp isn't wanted or needed.
         # At some point in the future we want to modify the HUP test suite to use the hostapp as the HUP target, rather than sending the DUT the docker image and doing a HUP using that file
         default: true
+      # The default is "auto" and `check-merge-tests` will determine whether or not to mark the hostapp as final.
+      # True values include case-insensitive true|yes|1|y
```
lmbarros, line 86: I am in a lifelong campaign against using 0 and 1 as Boolean literals, so I am obliged to ask if you really think this gives us any concrete benefit.

And if so, why using the C standard of "1 is true" instead of the shell standard of "0 is true"? ( <-- argument against interpreting 0 and 1 as Booleans disguised as an honest question.)

lmbarros (commented):
I left a small suggestion, but otherwise looks good!

.github/workflows/yocto-build-deploy.yml
klutchell: I agree, it felt icky as I was writing it.

I think we need to revisit this and find some other way to mandate that `this device type must never be finalized by this workflow` vs. `finalize if it makes sense to do so`

(P.S. bourne shell is probably wrong, I think most languages treat 1 as truthy, even the balena-supervisor!)

klutchell: How would you solve this particular problem @lmbarros ? If two inputs would be prone to having conflicting values, and one boolean input does not hold enough context?

@rcooke-warwick had another idea [here](https://balena.zulipchat.com/#narrow/stream/345889-balena-io.2Fos/topic/2024.2E7.20ESR/near/451502413)
> finalise-hostapp: true + test matrix defined + push event - check tests then deploy
> finalise-hostapp: true + dispatch event - then deploy without checking tests
> finalise-hostapp: false + test matrix defined + push event - don't deploy
> finalise-hostapp: false by default
> explicitly set finalise-hostapp: true for device types we want to deploy as final

lmbarros: Pretty much agree, @klutchell . I don't think there is a way to be really clean and clear *if* we want to offer this level of customization with different flags that interact in complex ways. The documentation of input options reading like an algorithm is a signal of that.

> This would be simpler if we removed the option to force finalize via workflow dispatch. Then the input could just be `finalize-on-push-if-tests-passed` and no one would be confused.

On a meta-level I love this. I mean, change the constraints (disallow force-finalize) in a way that allows for a much cleaner solution (one flag with one meaning and no(?) interactions with other inputs). I just feel I still don't get enough context to comment on whether we need force-finalize or not.

rcooke-warwick: personally I would prefer if we keep force-finalize, just because in the past we have had emergencies where we've had to semi-manually run the tests to release something for a customer. We shouldn't have had to do that, but it happened anyway and we needed to force a release.

rcooke-warwick: I agree though on how much cleaner it will be if that wasn't the case. 

`finalize-on-push-if-tests-passed: true + push` = deploy if tests passed
`finalize-on-push-if-tests-passed: true + dispatch` = don't check tests
`finalise-on-push-if-tests-passed: false` = never deploy no matter what

For device types where we don't want to deploy even with passed tests, we set it to false
For device types where we do, we set it to true
And we have it default to false to avoid accidents

Checking if test matrix was empty or not was done in an effort to avoid another input , but in the `finalize-on-push-if-tests-passed` , we wouldn't be avoiding another input. Sometimes greater clarity might be worth an extra input

klutchell: SGTM

klutchell: Couldn't we just allow building as draft on dispatch, and provide manual instructions for marking as final?

rcooke-warwick: thats also a good idea @klutchell

klutchell:
We are unlikely to go forward with this solution and there are new proposals [here](https://github.com/balena-os/balena-yocto-scripts/pull/366#discussion_r1688103146) and [here](https://balena.fibery.io/Work/Improvement/Update-finalize-hostapp-input-to-handle-never-2018).

PR #367, opened 2024-07-16
Fix handling of empty test matrix input

klutchell:
Currently the workflow breaks if no test matrix is provided. This PR was tested both with and without a test matrix.

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -296,8 +296,13 @@ jobs:
       # If the test_matrix is empty - it means there are no tests for the DT - so don't check tests, and don't finalise, unless manually done with "finalise-hostapp" input
       - name: Check test results
         # https://docs.github.com/en/actions/learn-github-actions/expressions#functions
-        # this expression checks if test matrix is any variation of empty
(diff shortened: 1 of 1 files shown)


.github/workflows/yocto-build-deploy.yml
```diff
@@ -977,7 +982,14 @@ jobs:
...
     environment: ${{ matrix.environment }}
-    if: ${{ github.event_name != 'push'}}
+    # https://docs.github.com/en/actions/learn-github-actions/expressions#functions
+    # this expression checks to make sure at least one test suite was provided via either matrix syntax
+    if: |
+      github.event_name != 'push' &&
+      (
+        join(fromJSON(inputs.test_matrix).test_suite) != '' ||
+        join(fromJSON(inputs.test_matrix).include.*.test_suite) != ''
+      )
```
lmbarros, line 992: IIUC, if we define a test matrix including `test_suite` but missing `environment`, the workflow would still break (because the test environment is required but not checked for).

klutchell: Correct, we don't fully prevent shooting ourselves in the foot but we at least do a small sanity of the input.

lmbarros: Sure, fair. The extra complexity of checking every corner case probably costs more than it is worth. I am fine with that!

PR #389, opened 2024-08-23
Add unique unique identifiers to summary job

klutchell:
The "All jobs" job is used for branch protection
as it captures the overall success/fail of all
the test matrices and build.

However without a unique name we are unable
to identify the device type in order to mark it as required.

This change adds a single instance matrix
to the build job and as a result will append the machine name and deploy environment in the job name.

Change-type: patch
Depends-on: https://github.com/balena-os/.github/pull/69

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -131,7 +131,7 @@ on:
 
 # https://docs.github.com/en/actions/using-jobs/using-concurrency
 concurrency:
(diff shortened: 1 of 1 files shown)


klutchell:
Before this change, note the `Yocto / All jobs` requirement
image

klutchell:
After this change, note the machine name and environment added to `Yocto / All jobs`
image

rcooke-warwick (approved):

PR #390, opened 2024-08-27
Fix if condition for checking merge commit test results

rcooke-warwick:
The logic for this step was broken - even when no test matrix was provided, it was chekcing for test results, which we don't want (as they won't exist)

Change-type: patch


klutchell:
LGTM

PR #392, opened 2024-08-29
use apt-get to install s4cmd instead of pip

rcooke-warwick:
using pip to install python packages also installable by debian is not recommended, and leads to errors while trying to do so. Switch to using apt-get to install s4cmd to address this

Change-type: patch


rcooke-warwick:
currently in draft as I've temporarily forced the s4cmd install to always happen, to test

rcooke-warwick:
tested successfully https://github.com/balena-os/balena-yocto-scripts/actions/runs/10614198822/job/29419594942?pr=392 

so unmarking as draft

klutchell (approved):

PR #396, opened 2024-08-30
only upload artifacts when test_matrix is defined

rcooke-warwick:
This prevents uploads of artifacts that will never be used - as they are only used in the tests step, if there are no tests then no need to upload

Change-type: patch


klutchell (approved):

rcooke-warwick:
blocked by https://github.com/balena-os/balena-generic/pull/516

PR #410, opened 2024-09-12
Encrypt balenaOS artifacts at rest in GitHub

ab77:
Applies symmetric encryption (PBKDF2 hardened) to balenaOS build assets prior to uploading them to GitHub for temporary storage between builds. Decrypts assets after downloading.

Requires: https://github.com/balena-os/.github/pull/83
Depends-on: https://github.com/balena-os/balena-generic/pull/525

https://balena.fibery.io/Work/Improvement/Encrypt-artifacts-used-between-jobs-2070

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -74,7 +74,7 @@ on:
         required: false
         type: string
         default: balena-cloud.com
-      # This input exists because we want the option to not auto-finalise for some device types, even if they have tests and those tests pass - for example some custom device types, the customer doesn't want new releases published until they green light it 
+      # This input exists because we want the option to not auto-finalise for some device types, even if they have tests and those tests pass - for example some custom device types, the customer doesn't want new releases published until they green light it
       finalize-on-push-if-tests-passed:
         description: Whether to finalize a hostApp container image to a balena environment, if tests pass.
         required: false
@@ -85,7 +85,7 @@ on:
         description: Force deploy a finalized release
         required: false
         type: boolean
-        default: false 
+        default: false
       deploy-ami:
         description: Whether to deploy an AMI to AWS
         required: false
@@ -278,14 +278,14 @@ jobs:
         run: |
           merge_commit=$(git rev-parse :/"^Merge pull request")
           echo "Found merge commit ${merge_commit}"
-          echo "merge_commit=${merge_commit}" >> "$GITHUB_OUTPUT"
+          echo "merge_commit=${merge_commit}" >>"${GITHUB_OUTPUT}"
 
       # This will control the deployment of the hostapp only - it will determine if it is marked as final or not
       # The hostapp being finalised is what determines if the API will present this OS version to user
       # If the test_matrix is empty - it means there are no tests for the DT - so don't check tests, and don't finalise, unless manually done with "force-finalize" input
       - name: Check test results
         # https://docs.github.com/en/actions/learn-github-actions/expressions#functions
-        # this expression checks that the test_matrix input is truthy - there is no test_matrix input provided in the device-repo workflow file, test results won't be checked, and 
+        # this expression checks that the test_matrix input is truthy - there is no test_matrix input provided in the device-repo workflow file, test results won't be checked, and
         # the release can't be finlized
         if: github.event_name == 'push' && inputs.test_matrix && inputs.finalize-on-push-if-tests-passed
         id: merge-test-result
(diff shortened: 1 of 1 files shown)


klutchell:
How can we make this conditional? Ideally we would only encrypt artifacts if certain conditions are met, like if the repository is public and sign image is enabled, for example.

Also please include a link to the Fibery improvement in the body of the PR.

.github/workflows/yocto-build-deploy.yml
```diff
@@ -1131,11 +1136,15 @@ jobs:
 
       - name: Decrypt artifacts
         working-directory: ${{ env.WORKSPACE }}
+        if: |
+          github.event.repository.public &&
+          inputs.sign-image == true
```
klutchell, line 1141: Rather than check the same conditions we used on upload, we could use the .enc extension only when some artifacts are encrypted, so this step is no-op if there are no files with that extension.

That makes it easier to change the conditions of encryption in the future, and even encrypt just a subset of the files, without it being a global on/off toggle.

ab77: Do we have a private repo, which is producing a signed image we can test this workflow on?

klutchell: Not yet, but we probably should. Let me open a PR we can use to test this PR.

(we need a public repo with signed images, so I'll get balena-generic ready to go)

In the meantime let's also add a condition to encrypt [private device types](https://github.com/balena-os/balena-yocto-scripts/blob/0684ef25f676baddd6c55a504b575776269588de/.github/workflows/yocto-build-deploy.yml#L172).

So `(inputs.sign-image == true || steps.balena-lib.outputs.is_private == 'true')` which would allow testing on balena-raspberrypi which is public with a mix of private and public device types.

klutchell: If that works we can merge without re-testing signed images IMO. At least then we will have the confidence to enable signed images for testing.

klutchell:
Testing signed images here: https://github.com/balena-os/balena-generic/pull/530

klutchell:
Testing private device-types on public repos [here](https://github.com/balena-os/balena-raspberrypi/actions/runs/10960106721/job/30433923393?pr=1171)

klutchell:
LGTM

klutchell:
@ab77 let's change the encryption key when we get a chance.

rcooke-warwick:
This is tested and working with both encrypt and decrypt here: https://github.com/balena-os/balena-raspberrypi/actions/runs/11126932142/job/30919311161?pr=1180

rcooke-warwick (approved):

.github/workflows/yocto-build-deploy.yml
```diff
@@ -1138,6 +1138,9 @@ jobs:
           path: ${{ env.WORKSPACE }}
 
       - name: Decrypt artifacts
+        if: |
+          github.event.repository.private != true &&
+          (inputs.sign-image == true || needs.build.outputs.is_private == 'true')
```
klutchell, line 1143: @rcooke-warwick I asked Anton to remove this so we had one fewer place where conditions were used to decide between encryption and decryption. Like if we changed the conditions for encryption and forgot to update the decryption conditions, it might take a while to notice we broke it.

Was it not correctly detecting the lack of *.enc files?

rcooke-warwick: Thats right, 

```
Run for artifact in *.enc **/*.enc; do
+ for artifact in *.enc **/*.enc
+ openssl enc -v -d -aes-256-cbc -k *** -pbkdf2 -iter 310000 -md sha256 -salt -in '*.enc' -out '*'
bufsize=8192
Can't open "*.enc" for reading, No such file or directory
40D756EFD17F0000:error:80000002:system library:BIO_new_file:No such file or directory:../crypto/bio/bss_file.c:67:calling fopen(*.enc, rb)
40D756EFD17F0000:error:10000080:BIO routines:BIO_new_file:no such file:../crypto/bio/bss_file.c:75:
```

rcooke-warwick: If we're too lenient about the logic in this step passing, we risk scenarios like:

- we break the encryption step
- we break the decryption step 
- we change the logic to .enc files get written to a different place, or are called something else

And then trying to flash the DUT with whatever, or getting as far as the tests, which will then fail - but its harder to ID where the failure happened if it gets that far. Just to keep in mind

klutchell: Ack, sounds good

PR #412, opened 2024-09-18
Expose secure boot testing features via matrix input

klutchell:
Some device types (generic-armd64) need to run the test
suites twice for the same base image, once with secure boot
vars enabled in Leviathan, and once without.

Rather than run the suites in series, or require building the
image twice, we can allow the test matrix to define separate runs
for the two modes.

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -103,6 +103,7 @@ on:
       # - worker_fleets: The testbot fleets for finding available Leviathan workers. Not used for QEMU workers. Can accept a list of apps separated by commas, no spaces in between
       # - test_org: The organization to use for testing cloud functionality. This default org is `testbot`
       # - runs_on: A JSON array of runner labels to use for the test job(s). For qemu workers use the labels `["self-hosted", "X64", "kvm"]`.
+      # - secure_boot: (truthy) Enable secure boot testing flags QEMU_SECUREBOOT=1 and FLASHER_SECUREBOOT=1. Default is false.
       # To use specific settings for each test job, create an include array like this...
       # {"include": [
       #   {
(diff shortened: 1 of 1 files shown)


rcooke-warwick:
added 

```
QEMU_INTERNAL_STORAGE=0
QEMU_MEMORY=4G
```

To environment based off https://balena.zulipchat.com/#narrow/stream/345889-balena-io.2Fos/topic/Secure.20boot.20and.20FDE.20check-in/near/433240205

rcooke-warwick:
https://github.com/balena-os/leviathan/blob/master/action.yml#L46

There is a conflict here - with the SB matrix it tries to make artifacts with the same name amd fails

vipulgupta2048:
@rcooke-warwick Do you have a run where the conflict appears. With Kyle's change, it seems we won't be doing a normal run when secureboot will be enabled. Hence, yes the artifacts could be better named to exactly show the secureboot run but there won't be other artifacts present?

rcooke-warwick:
There are other artifacts present because there is a SB and non SB version of the tests run, and both produce artifacts with the same name 

E.g https://github.com/balena-os/balena-generic/actions/runs/11216427427/job/31184440638?pr=530

rcooke-warwick:
We also still need to figure out why the SB enables DUT never comes online , even after the 4G memory being enabled

rcooke-warwick:
> > There are other artifacts present because there is a SB and non SB version of the tests run, and both produce artifacts with the same name
> > E.g https://github.com/balena-os/balena-generic/actions/runs/11216427427/job/31184440638?pr=530
> > ```
> > >Run actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874
> > With the provided path, there will be 4 files uploaded
> > Artifact name is valid!
> > Root directory input is valid!
> > Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We need another variable in the artifact name to differentiate it
> 
> This fix will have to be in the Leviathan workflow to generate unique artifact names -- ideally unique with a bunch of properties, like machine, signed/unsigned, worker type, etc. @vipulgupta2048

https://github.com/balena-os/leviathan/pull/1243/ here is a shot at that

rcooke-warwick (approved):

vipulgupta2048 (approved):

PR #415, opened 2024-09-19
Give the build job a unique name with the machine and environment

klutchell:
This allows us to mark specific build-only jobs as required.

Specifically for cases where tests are defined, but those tests may not be marked as required for an ESR branch to merge, so marking 'Yocto / All jobs' as required for that device is not ideal.

Change-type: patch

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -159,6 +159,13 @@ jobs:
     runs-on: ${{ fromJSON(inputs.build-runs-on) }}
     environment: ${{ inputs.deploy-environment }}
 
+    strategy:
+      fail-fast: true
+      matrix:
+        include:
+          - machine: ${{ inputs.machine }}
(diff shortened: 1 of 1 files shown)


rcooke-warwick:
To check I understand right, this is for the case where we have something in a device repo that has tests defined, but we're not deploying finalised ESR releases for - for example:

- some of the `balena-raspberrypi` repo device types: `raspberrypi0-2w`
- `generic-amd64-fs` in `balena-generic`

But on an ESR branch, these aren't required - so a PR might merge without tests passing for these device types - but if that happens then a finalised ESR release won't be deployed, only a draft - isn't that what we want?

What would be worse was if tests DID pass, because then an ESR release would be deployed for a DT we don't guarantee ESR support for? This is why we commented out the ESR regexp on the branch checks for non ESR devices isn't it?, e.g: https://github.com/balena-os/balena-intel/blob/master/.github/workflows/surface-pro-6.yml#L11  If we offered ESR for all tested devices this doesn't matter of course.

PR #417, opened 2024-09-26
use genericx86-64-ext image for testing the workflow

rcooke-warwick:
Using genericx86-64-ext as its smaller, faster to build and the repo is blocked less often.

In this particular case the `balena-generic` repo is blocked due to an upstream change breaking the HUP suite on secure boot devices. This is a problem as we need to merge other meta-balena fixes to fix builds for `balena-generic` - using `balena-intel` , already working with GHA workflows unblocks fixes to this workflow.

We don't actually care about the device type being testedl, just that the workflow is tested

Change-type: patch


vipulgupta2048 (approved):

klutchell:
LGTM

PR #419, opened 2024-09-27
CI: Update checkout settings

Page-:
Change-type: patch


rcooke-warwick:
same as https://github.com/balena-os/balena-yocto-scripts/pull/416 

we were just blocked by https://github.com/balena-os/balena-yocto-scripts/pull/417 until 5 mins ago

PR #421, opened 2024-10-01
alter trigger in test workflow

rcooke-warwick:
Change-type: patch


rcooke-warwick:
no longer needed after https://github.com/balena-os/balena-yocto-scripts/pull/436

PR #441, opened 2024-10-29
add pull_request: write permissions to default token

rcooke-warwick:
For use with comment based workflow run approvals

Change-type: patch


rcooke-warwick:
same as https://github.com/balena-os/balena-yocto-scripts/pull/442  , closing

PR #449, opened 2024-11-16
github/workflows/yocto-build-deploy.yml: Add hostapp metadata to OS r…

floion:
…elease

This is needed for supervisor managed OS updates. For example, the supervisor can check this metadata against the equivalent field from /etc/os-release to determine if the device is already running the latest OS release.

Change-type: patch

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -382,6 +382,9 @@ jobs:
           meta_balena_version="$(balena_lib_get_meta_balena_base_version)"
           echo "meta_balena_version=${meta_balena_version}" >>"${GITHUB_OUTPUT}"
 
+          device_repo_revision="$(git rev-parse --short HEAD)"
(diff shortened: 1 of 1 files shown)


.github/workflows/yocto-build-deploy.yml
```diff
@@ -814,9 +818,21 @@ jobs:
...
+version: '2.4'
+services:
+  hostapp:
+    image: ${_local_image}
+    labels:
+      io.balena.image.store: 'root'
+      io.balena.image.class: 'hostapp'
+      io.balena.update.requires-reboot: '1'
+      io.balena.private.hostapp.board-rev: '${DEVICE_REPO_REV}'
+EOF
```
klutchell, line 832: This indentation needs to be aligned with the level of the other commands above like `cat > ...` in order to be parsed as valid yaml. When it is executed it will still behave as expected and not have any leading whitespace in the compose file.

PR #455, opened 2024-11-29
Use app installation tokens instead of default github token

rcooke-warwick:
Change-type: patch

### DIFF
--- a/.github/workflows/genericx86-64-ext.yml
+++ b/.github/workflows/genericx86-64-ext.yml
@@ -31,6 +31,9 @@ on:
 jobs:
   yocto:
     name: Yocto
+    permissions:
+      # Have to explicitly get permission here as not yet supported by gh app tokens
+      id-token: write # This is required for requesting the JWT #https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services#requesting-the-access-token
     uses: ./.github/workflows/yocto-build-deploy.yml
     # Prevent duplicate workflow executions for pull_request (PR) and pull_request_target (PRT) events.
     # Both PR and PRT will be triggered for the same pull request, whether it is internal or from a fork.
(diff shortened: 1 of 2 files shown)


.github/workflows/yocto-build-deploy.yml
```diff
@@ -213,6 +212,28 @@ jobs:
...
+          installation_retrieval_mode: organization
+          installation_retrieval_payload: ${{ github.repository_owner }}
+          private_key: ${{ secrets.GH_APP_PRIVATE_KEY }}
+          # actions:read - We are fetching workflow run results of a merge commit when workflow is triggered by new tag, to see if tests pass
+          # pull-requests:write - # Read is required to fetch the PR that merged, in order to get the test results. Write is required to create PR comments for workflow approvals.
+          # conents:read - required for cloning private submodules
+          permissions: |-
+            {
+              "contents": "read",
+              "actions": "read",
```
klutchell, line 233: The Flowzone app does not have `actions: read` IIRC, and neither does the new defaults for the automatic GITHUB_TOKEN. I think we need to decide which is better to use here.

.github/workflows/genericx86-64-ext.yml
```diff
@@ -31,6 +31,9 @@ on:
 jobs:
   yocto:
     name: Yocto
+    permissions:
+      # Have to explicitly get permission here as not yet supported by gh app tokens
+      id-token: write # This is required for requesting the JWT #https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services#requesting-the-access-token
```
klutchell, line 36:
```suggestion
      actions: read
      contents: read
      packages: read
      id-token: write # This is required for requesting the JWT #https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services#requesting-the-access-token
      pull-requests: write
```

klutchell:
Superseded by https://github.com/balena-os/balena-yocto-scripts/pull/458

PR #462, opened 2024-12-12
Support using any org administrator account for the hostApp creation

thgreasi:
Change-type: patch


klutchell (changes requested):
Thanks for opening this @thgreasi , but we are actually planning to move this logic outside of the CI workflow an into something that can be run manually. I would prefer the workflow that runs on every repo not have access to an administrator account.

rcooke-warwick:
deleting hostapp creation from workflow here: https://github.com/balena-os/balena-yocto-scripts/pull/468 So closing for now

PR #465, opened 2024-12-13
Disable tests on workflow dispatch trigger

rcooke-warwick:
Workflow dispatch is just used for manual deploys - we don't want tests to run after.

Change-type: patch


klutchell (changes requested):
Sometimes we do want to run tests on workflow dispatch, like when Alex needed to run secure boot tests that are not enabled in meta-balena.

An alternate approach would be to expose the test matrix input to the workflow_dispatch call.

PR #467, opened 2024-12-13
Compress artifacts with gzip and zstd before encrypting

klutchell:
Otherwise encrypted images are unable to be compressed. This saves a lot of space and time on uploading/downloading artifacts.

Compressing images before upload makes a big difference because a bunch of the image space is zeros. In testing I saw the compressed artifacts were only 26% the size of the uncompressed artifacts.

With encryption forced, the artifacts are ~1.2GB on this PR. Previously encrypted artifacts for generic-amd64 were ~4.6GB 🎉


klutchell:
With encryption forced, the artifacts are ~1.2GB on this PR. Previously encrypted artifacts for generic-amd64 were ~4.6GB.

rcooke-warwick (approved):

ab77 (approved):

PR #474, opened 2024-12-18
workflows: yocto-build-deploy: add signing keys arguments

alexgg:
Change-type: patch


klutchell:
Before merging this, we should probably [define the variable in the repository](https://github.com/github/safe-settings/blob/main-enterprise/docs/github-settings/2.%20repository-variables.md) for safe-settings to apply.

So you'll need a top-level variables property [here](https://github.com/balena-os/.github/blob/master/.github/repos/balena-iot-gate-imx8.yml).

.github/workflows/yocto-build-deploy.yml
```diff
@@ -457,12 +457,13 @@ jobs:
           SIGN_API_KEY: "${{ secrets.SIGN_API_KEY }}"
           SIGN_GRUB_KEY_ID: 2EB29B4CE0132F6337897F5FB8A88D1C62FCC729
           SIGN_KMOD_KEY_APPEND: "${{ secrets.SIGN_KMOD_KEY_APPEND }}"
+          SIGN_HAB_PKI_ID: "${{ vars.SIGN_HAB_PKI_ID || '6d74b15cbc5df27fdc8d470a7c71edb3' }}"
         run: |
           BARYS_ARGUMENTS_VAR="${BARYS_ARGUMENTS_VAR} -a SIGN_API=${SIGN_API}"
           BARYS_ARGUMENTS_VAR="${BARYS_ARGUMENTS_VAR} -a SIGN_API_KEY=${SIGN_API_KEY}"
           BARYS_ARGUMENTS_VAR="${BARYS_ARGUMENTS_VAR} -a SIGN_GRUB_KEY_ID=${SIGN_GRUB_KEY_ID}"
```
klutchell, line 464: Do we want to take this opportunity to define SIGN_GRUB_KEY_ID as a var as well?

alexgg: sure, it's done @klutchell

.github/workflows/yocto-build-deploy.yml
```diff
@@ -585,6 +595,9 @@ jobs:
 
           cat "${AUTO_CONF_FILE}"
 
+          >&2 eval "$(ssh-agent)"
+          echo "${{ secrets.YOCTO_SSH_PRIVATE_KEY_B64 }}" | base64 -d | ssh-add - >&2
```
jakogut, line 600: [PEM formatted private keys](https://en.wikipedia.org/wiki/Privacy-Enhanced_Mail) like those used by OpenSSH are already base64 encoded, with a header and footer wrapping the encoded data.

These labels can be stripped and the base64 content passed as the secret, and they can be reinserted before passing the key to `ssh-add`.

This mirrors the approach used with `SIGN_KMOD_KEY_APPEND`. https://github.com/balena-os/meta-balena/pull/3114/files

jakogut (changes requested):

.github/workflows/yocto-build-deploy.yml
klutchell: The problem is the header and the footer and newlines are not suitable for storing as Git secrets unless we update them by hand. What is the best format to retain the secret as a single one-line string?

alexgg: @jakogut why is that preferred?

alexgg:
> Before merging this, we should probably [define the variable in the repository](https://github.com/github/safe-settings/blob/main-enterprise/docs/github-settings/2.%20repository-variables.md) for safe-settings to apply.
> 
> So you'll need a top-level variables property [here](https://github.com/balena-os/.github/blob/master/.github/repos/balena-iot-gate-imx8.yml) or [here](https://github.com/balena-os/.github/blob/master/.github/settings.yml).
> 
> I guess you have the default in place, so we don't have to wait. Is that default meant to apply to all device repos?

The settings have already been extended with the signing environment that contain the top level settings.

.github/workflows/yocto-build-deploy.yml
alexgg: @jakogut not sure if this is just a conversation about subjective preference or there are actual advantages - would you mind to chip in so we can move this PR forward?

jakogut: @alexgg It's not a showstopper, I was just pointing out that the PEM content is already base64 encoded, and the header and footer are a known quantity. We don't need to base64 encode data that's already base64 encoded.

klutchell:
@alexgg if this works for your use case, then it looks good to me after a rebase

.github/workflows/yocto-build-deploy.yml
klutchell: @jakogut I think what we have is fine, and makes it simpler for Ops to rotate the keys without knowing what is okay to strip and what has to be kept intact, or running a script. They are quite familiar with piping secrets to base64 on Windows.

alexgg: This was actually merged on an individual PR that this branch had rebased - https://github.com/balena-os/balena-yocto-scripts/pull/479.

I understand we encode content that is already encoded, but as Kyle says by doing that we avoid having to mangle the contents so I too prefer it. It might be different if this was on device instead of just the build pipeline.

klutchell (approved):
image
This is not a valid value for change-type, but once fixed it LGTM

jakogut (approved):

PR #479, opened 2024-12-23
workflows: yocto-build-deploy: add SSH key to environment

alexgg:
This allows the build step to access private repositories via SSH.

Change-type: patch

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -38,6 +38,9 @@ on:
       YOCTO_CACHE_SECRET_KEY:
         description: "Self-hosted runner S3 secret key for the yocto-svcacct user."
         required: false
+      YOCTO_SSH_PRIVATE_KEY_B64:
+        description: "SSH key to access balena-os private repositories."
+        required: false
 
     inputs:
       build-runs-on:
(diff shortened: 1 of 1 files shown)


.github/workflows/yocto-build-deploy.yml
```diff
@@ -585,6 +595,9 @@ jobs:
 
           cat "${AUTO_CONF_FILE}"
 
+          >&2 eval "$(ssh-agent)"
+          >&2 echo "${YOCTO_SSH_PRIVATE_KEY_B64}" | base64 -d | ssh-add -
```
klutchell, line 599:
```suggestion
          >&2 echo "${{ secrets.YOCTO_SSH_PRIVATE_KEY_B64 }}" | base64 -d | ssh-add -
```
This way we can avoid the secret being exposed to the build container as an env var, and limit the scope a bit.

alexgg: Testing right now.

klutchell:
@alexgg did this work in your testing (it was able to clone the private yocto sources)?

alexgg:
> @alexgg did this work in your testing (it was able to clone the private yocto sources)?

It did work @klutchell the test run is https://github.com/balena-os/balena-iot-gate-imx8/actions/runs/12669323822/job/35317948384, and the build job completed fine.

klutchell (approved):

PR #487, opened 2025-01-07
Upload yocto build logs with existing artifacts

klutchell:
Change-type: patch

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -599,6 +599,23 @@ jobs:
             exit 1
           fi
 
+      # We don't need to encrypt these as they are not sensitive if the repo is public anyway.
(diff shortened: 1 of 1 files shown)


vipulgupta2048 (commented):
LGTM, similar setup in leviathan here, probably can add compression and retention settings too, https://github.com/balena-os/leviathan/blob/6652ce020ed9c242914bb97bfd9852f4be1f77da/action.yml#L56

vipulgupta2048 (approved):
Verified and work, created a 31.9 MB package and uploaded. Doesn't show up on the UI yet but maybe it does later after merge. 
Screenshot 2025-01-09 at 2 26 56 PM


Just wanted to make sure if we need all these log files too to debug the build.

klutchell:
@vipulgupta2048 thanks for checking, those are indeed the files we need. Confirmed with Alex.

PR #496, opened 2025-01-23
barys: build unique machine-image combinations

alexgg:
Keep track of the different machine-image combos being built not to duplicate fields.

This allows to have multiple device types that build the same machine, for example if they have different feature layers.

Change-type: patch

### DIFF
--- a/build/barys
+++ b/build/barys
@@ -253,7 +253,11 @@ while [[ $# -ge 1 ]]; do
             if [ -z "$2" ]; then
                 log ERROR "\"$1\" argument needs a value."
             fi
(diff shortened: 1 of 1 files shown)


build/barys
```diff
@@ -578,23 +582,29 @@ else
           continue
         fi
         if [ -z "$BITBAKE_TARGET" ]; then
-            IMAGE=`jq -r '.yocto | select(.machine == '\"${machine}\"').image' $json`
```
alexgg, line 581: This is redundant as the machine check is carried out above.

klutchell (approved):

PR #520, opened 2025-03-02
Unpack all artifacts for S3 upload

klutchell:
We need most of the files from the root of the artifacts directory for the S3 upload, not just the files in the image directory.

See: https://balena.fibery.io/Security/Information_Security_and_Reliability_Incident/Some-finalized-OS-releases-were-missing-S3-files-62


klutchell:
Test run with with `force-finalize: true`: https://github.com/balena-os/balena-yocto-scripts/actions/runs/13618732194/job/38065723812?pr=520

floion (approved):

PR #555, opened 2025-03-21
Leviathan: Allow specifying alternative image to test

rcooke-warwick:
This is for use in at least 3 device types - where the customer requires a non-flasher image to be available by default via downloads, but for testing purposes we need ot use the flasher image. This change allows for selecting the flasher image using the test_matrix input - i.e `image: balena-flasher.img`

Change-type: patch

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -128,6 +128,7 @@ on:
       # - test_org: The organization to use for testing cloud functionality. This default org is `testbot`
       # - runs_on: A JSON array of runner labels to use for the test job(s). For qemu workers use the labels `["self-hosted", "X64", "kvm"]`.
       # - secure_boot: (truthy) Enable secure boot testing flags QEMU_SECUREBOOT=1 and FLASHER_SECUREBOOT=1. Default is false.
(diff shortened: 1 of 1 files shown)


.github/workflows/yocto-build-deploy.yml
```diff
@@ -1944,6 +1944,8 @@ jobs:
       WORKER_TYPE: ${{ matrix.worker_type || 'testbot' }}
       BALENACLOUD_APP_NAME: ${{ matrix.worker_fleets || 'balena/testbot-rig,balena/testbot-rig-partners,balena/testbot-rig-x86,balena/testbot-rig-partners-x86' }}
       BALENACLOUD_ORG: ${{ matrix.test_org || 'testbot' }}
+      # Allows for specifying an alternative build artifact to the default
+      BALENA_OS_IMAGE: ${{ matrix.image || 'balena' }}
```
klutchell, line 1948: This is just to allow switching between the flasher and the raw image?

It doesn't appear to allow any changing of the path or the file extension, just the image _basename_, so I think we should rename the matrix property and comment it well, including as part of the matrix description at the top of the workflow.

rcooke-warwick: Its for switching between flasher vs raw in the case where the default `balena.img` artifact is the raw image (downloadable via dashboard) but for the purposes of testing we need to use a flasher image (edge case for several CDS device types). 

Originally I allowed specifying the full name, not just the base name, changed it because of an error I was getting and was trying out several fixes - will revert to the full `balena.img` / `balena-flasher.img` . Agreed on adding extra description and using a better name

klutchell (approved):

PR #557, opened 2025-03-24
workflows: yocto-build-deploy: setup AWS authentication for tests

alexgg:
The secure boot test for signed kernel modules, when run for private device types, needs to be able to download kernel headers from an authenticated S3 bucket, at least until the assets are moved to using web resources.

Change-type: patch


alexgg:
The kernel file artifact is already locally available - no need to pull again.

PR #565, opened 2025-03-27
AMI deploy: parameterise test org + fleet for AMI test

rcooke-warwick:
Allows using vars to set the org + fleet name for use in AMI tests, instead of hard coding and dynamically creating fleets.

This is to make management of permissions simpler

Change-type: patch

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -1363,7 +1363,8 @@ jobs:
   ami-deploy:
     name: Deploy AMI
     runs-on: ${{ fromJSON(inputs.build-runs-on) }}
-    if: inputs.deploy-ami && (github.event_name == 'push' || github.event_name == 'workflow_dispatch')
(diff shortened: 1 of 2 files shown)


.github/workflows/yocto-build-deploy.yml
```diff
@@ -1649,32 +1649,34 @@ jobs:
...
-        id: ami-test-fleet
+      - name: Setup AMI test Device
+        id: ami-test-device
         env:
           HOSTOS_VERSION: "${{ needs.balena-lib.outputs.os_version }}"
-          AMI_TEST_ORG: testbot
           AMI_TEST_DEV_MODE: true
+          # Pre-made permanent fleet for provisioning AMI's in to for testing
+          AMI_TEST_ORG: ${{ vars.AMI_TEST_ORG }}
+          AMI_TEST_FLEET: ${{ vars.AMI_TEST_FLEET }}
```
klutchell, line 1659: Will it help to test this PR to codify these vars first, or just set some sane defaults here?

PR #597, opened 2025-04-22
Fix passing the BALENA_API_DEPLOY_KEY is "Check if device-type is private" step

thgreasi:
Change-type: patch
See: https://github.com/actions/github-script/issues/253#issuecomment-1165911869

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -384,13 +384,14 @@ jobs:
           API_ENV: ${{ env.BALENARC_BALENA_URL }}
           TRANSLATION: "v6"
           DEVICE_SLUG: ${{ steps.balena-lib.outputs.device_slug }}
+          BALENAOS_TOKEN: ${{ secrets.BALENA_API_DEPLOY_KEY }}
         with:
           result-encoding: json
           script: |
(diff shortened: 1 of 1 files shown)


.github/workflows/yocto-build-deploy.yml
```diff
@@ -384,13 +384,14 @@ jobs:
           API_ENV: ${{ env.BALENARC_BALENA_URL }}
           TRANSLATION: "v6"
           DEVICE_SLUG: ${{ steps.balena-lib.outputs.device_slug }}
+          BALENAOS_TOKEN: ${{ secrets.BALENA_API_DEPLOY_KEY }}
```
thgreasi, line 387: Afaict that's also how we do it in scripts using curls
https://github.com/balena-os/balena-yocto-scripts/blob/acf945c2b6b9a574bcf3927288805474434e44d1/.github/workflows/yocto-build-deploy.yml#L985-L1033C205

.github/workflows/yocto-build-deploy.yml
```diff
@@ -384,13 +384,14 @@ jobs:
...
         with:
           result-encoding: json
           script: |
             const result = await fetch(`https://api.${process.env.API_ENV}/${process.env.TRANSLATION}/device_type?\$filter=slug%20eq%20%27${process.env.DEVICE_SLUG}%27&\$select=slug,is_private`, {
               headers: {
                 'Content-type': 'application/json',
-                'Authorization': `Bearer ${{ secrets.BALENA_API_DEPLOY_KEY }}`
```
klutchell, line 393: The substitution here would happen before the script runs, so both versions would be valid.

PR #638, opened 2025-05-22
Split artifact uploads into hostapp and testing artifacts

klutchell:
This will improve the download time for each, and allow
reduction of required hostapp files without breaking tests.

Moving the creation of deflate files to the build job enables us to upload S3 artifacts in both the deploy-S3 and the deploy-hostapp steps (as webresources).

See: https://balena.fibery.io/Work/Improvement/Prepare-deflate-files-before-hostapp-deploy-2913

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -524,6 +524,8 @@ jobs:
       AUTO_CONF_FILE: "${{ github.workspace }}/build/conf/auto.conf"
       BALENARC_BALENA_URL: ${{ needs.balena-lib.outputs.balena_host }}
       DEPLOY_PATH: ${{ github.workspace }}/deploy
+      # Note that this is the DEPLOY_PATH + the device slug and os version
+      S3_DEPLOY_PATH: "${{ github.workspace }}/deploy/${{ needs.balena-lib.outputs.device_slug }}/${{ needs.balena-lib.outputs.os_version }}"
 
     defaults:
       run:
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -892,15 +892,23 @@ jobs:
 
           find "${DEPLOY_PATH}" -exec ls -lh {} \;
 
-      # Encrypt .img and .img.zip files and remove the originals
       # Encryption is required for private device types, and signed images.
+      # We do this to obfuscate private device images on public repos,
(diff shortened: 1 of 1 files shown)


.github/workflows/yocto-build-deploy.yml
```diff
@@ -850,41 +850,61 @@ jobs:
...
         if: inputs.sign-image || needs.balena-lib.outputs.is_private == 'true'
         env:
-          ARTIFACTS_TAR: "${{ runner.temp }}/artifacts.tar.zst"
-          ARTIFACTS_ENC: "${{ runner.temp }}/artifacts.tar.zst.enc"
           PBDKF2_PASSPHRASE: ${{ secrets.PBDKF2_PASSPHRASE }}
         run: |
-          openssl enc -v -e -aes-256-cbc -k "${PBDKF2_PASSPHRASE}" -pbkdf2 -iter 310000 -md sha256 -salt -in "${ARTIFACTS_TAR}" -out "${ARTIFACTS_ENC}"
-          rm "${ARTIFACTS_TAR}"
+          set -x
+          for file in "${S3_DEPLOY_PATH}"/**/*.img "${S3_DEPLOY_PATH}"/**/*.img.zip "${S3_DEPLOY_PATH}"/**/*.docker "${S3_DEPLOY_PATH}"/**/*.deflate; do
```
rcooke-warwick, line 903: extreme nit - but is there anything that is not the images / hostapp files that should be encrypted? I don;t think there is, just sanity checking - the manifest, kernel module headers etc?

.github/workflows/yocto-build-deploy.yml
```diff
@@ -982,18 +981,19 @@ jobs:
...
+      DEPLOY_PATH: ${{ github.workspace }}/deploy
 
     steps:
 
       # Use gh to download the build artifacts from the build job
       # https://github.com/actions/download-artifact/issues/396#issuecomment-2796144940
       # https://cli.github.com/manual/gh_run_download
       # https://cli.github.com/manual/gh_help_environment
-      - name: Download build artifacts
+      - name: Download hostapp artifacts
```
rcooke-warwick, line 994: Nit: Given that the hostapp artifacts are a superset of the testing artifacts, do we gain some speed by taking the testing artfiacts set out of the hostapp artifacts now, instead of later - and then downloading the testing artifacts as well as hostapp artifacts in the "deploy hostapp" job?

It might not be worth the gains from not uploading the same thing twice , so just a question

PR #681, opened 2025-06-18
prepare deflate artifacts even on PRs

rcooke-warwick:
On PRs if there are no deflate files generated, the subsequent encrypt step will fail (it is run only for private DTs). Rather than add any further complication to that step, its cleaner to just prepare the deflate files regardless. Especially as we might be adding them to PR draft releases in the future anyway. The time cost is low for this step anyway.

Change-type: patch


klutchell (approved):
LGTM

PR #724, opened 2025-07-28
Dedupe raw or flasher images files during artifact prepare

klutchell:
Currently generic device types are uploading 3 images in the test artifacts

balena.img.enc
balena-flasher.img.enc
balena-raw.img.enc

But balena.img is always going to match either the flasher or the raw image depending on the device spec.

This change will replace balena.img(.enc) with a symlink when possible in only the test artifacts to reduce the upload and download size.

Needs to retain support for https://github.com/balena-os/balena-yocto-scripts/pull/640

Testing with generic (MBR) went from 2.68GB to 1.95GB for testing artifacts.

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -1529,6 +1529,13 @@ jobs:
             unzip "${zip}" -d "$(dirname "${zip}")"
           done
 
+          # If balena.img is a symlink, replace it with the file it points to.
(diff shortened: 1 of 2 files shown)


klutchell:
> does this affect hostapp-artifacts?

It shouldn't, no. It intentionally only touches the zip files, not the img files which are used to generate deflates and included in the hostapp artifacts.

We should compare the artifacts before and after to be sure.

klutchell:
Superseded by https://github.com/balena-os/balena-yocto-scripts/pull/726

PR #725, opened 2025-07-28
upload test artifacts: don't upload all image variants

rcooke-warwick:
We only need balena.img, as that is the artifact used in the test. This is to save large amounts of bandwidth for device types such as generic-amd64 which have 3 variants.

Change-type: patch


klutchell:
Superseded by https://github.com/balena-os/balena-yocto-scripts/pull/724

PR #726, opened 2025-07-29
Split testing artifacts into individual files

klutchell:
This allows downstream jobs to only download the files they need, and one at a time, avoiding issues with large artifact download timeouts.

Change-type: minor


rcooke-warwick (commented):
this works on https://github.com/balena-os/balena-generic/actions/runs/16627112405/job/47819611756?pr=632 and anecdotally, I haven't seen any download artifact fails - we should probably just double check the deploys still work, then LGTM

klutchell:
@rcooke-warwick I've tested the deploy [here](https://github.com/balena-os/balena-generic/actions/runs/16966668777/job/48195917516?pr=632), and compared the S3 files with a previous deploy and the file list is the same.

The hostapp deploy to staging failed because the rev is out of sync, but it loaded the docker image just fine so the deploy would have worked.

As far as I'm concerned this is ready for merge so we can start to reap the benefits as soon as possible.

PR #895, opened 2026-04-17
Add hostapp composition build pipeline

klutchell:
Build a balenaOS hostapp from a Docker Compose composition declared by
the device repo. Services declaring x-build are built in parallel via
a matrix job; outputs and registry images are assembled into a single
release at deploy time.

See docs/specs/hostapp-composition.md for the contract.

See: https://balena.fibery.io/Work/Improvement/Extensions-build-pipeline-changes-4332

Change-type: minor

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -270,6 +270,9 @@ jobs:
       is_esr: ${{ startsWith(github.ref, 'refs/tags/v20') }}
       deploy_path: ${{ github.workspace }}/deploy/${{ steps.balena-lib.outputs.device_slug }}/${{ steps.balena-lib.outputs.os_version }}
       balena_host: ${{ env.BALENARC_BALENA_URL }}
+      build_services: ${{ steps.build-matrix.outputs.services }}
+      build_matrix: ${{ steps.build-matrix.outputs.matrix }}
+      has_extensions: ${{ steps.build-matrix.outputs.has_extensions }}
 
     steps:
       # Generate an app installation token that has access to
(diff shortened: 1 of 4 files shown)


.github/workflows/yocto-build-deploy.yml
```diff
@@ -377,6 +379,134 @@ jobs:
...
+          if [ -f "${META_COMPOSITION}" ] && [ -f "${COMPOSITION}" ]; then
+            echo "[INFO] Merging meta-balena composition with device repo overlay"
+            yq eval-all 'select(fileIndex == 0) *d select(fileIndex == 1)' \
+              "${META_COMPOSITION}" "${COMPOSITION}" > "${COMPOSITION}.tmp"
+            mv "${COMPOSITION}.tmp" "${COMPOSITION}"
+          elif [ -f "${META_COMPOSITION}" ]; then
+            echo "[INFO] Using meta-balena composition (no device repo overlay)"
+            cp "${META_COMPOSITION}" "${COMPOSITION}"
+          elif [ ! -f "${COMPOSITION}" ]; then
+            echo "[INFO] No composition file found, generating default"
```
rcooke-warwick, line 397: If there is no composition file (we're building from an old tag, the composition file not merged into the device repo yet, or its an ESR release cut from before the composition file is merged into the device repo) - this will build in the legacy style? Yocto scripts master action will be backwards compatible?

.github/workflows/yocto-build-deploy.yml
```diff
@@ -377,6 +379,134 @@ jobs:
...
+          # A device overlay opts out of a service entirely by nulling the whole key (e.g., `kernel-modules: ~`).
+          # yq's *d operator propagates the null through the merge; this step then deletes the null entries.
+          # For build-only opt-out (keep the service in deploy but drop it from the build matrix), a device overlay
+          # sets x-build: ~ alongside a concrete image: field — same null-override pattern, applied to a nested key.
+          yq -i '.services |= with_entries(select(.value != null))' "${COMPOSITION}"
+
+          echo "[INFO] Composition:"
+          cat "${COMPOSITION}"
+
+      - name: Validate composition
```
rcooke-warwick, line 422: quite a nit, but was wondering if there's soemthing akin to a json schema to more concisely declare what a valid composition looks like.

.github/workflows/yocto-build-deploy.yml
rcooke-warwick, line 977: ignore if I'm not understanding right - but this step is the yocto build , should it run if the image class is not a hostapp?

.github/workflows/yocto-build-deploy.yml
```diff
@@ -1209,7 +1375,7 @@ jobs:
       # https://github.com/actions/upload-artifact
       - name: Upload hostapp artifacts
         uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
-        if: github.event_name == 'push' || github.event_name == 'workflow_dispatch' || inputs.force-finalize
+        if: matrix.image_class == 'hostapp'
```
rcooke-warwick, line 1403: this means we are uploading hostapp artifacts even when this is a PR triggered run - which we weren't doing before, as we currently aren't using them in tests. I can't remember why exactly, it was maybe cost related? - is there a motivation for this change?

rcooke-warwick: Ignore - I'm guessing this is because we need to fetch them in a later step to create the final image with hostapp + extensions?

alexgg (approved):
lgtm

.github/workflows/yocto-build-deploy.yml
klutchell: Correct!

.github/workflows/yocto-build-deploy.yml
klutchell: IIRC is was just to shave a few seconds off the build job, and save on storage. But I think that was over-engineered for what we actually need, these artifacts all expire in 1 day anyway and it's helpful to always generate them for debugging.

.github/workflows/yocto-build-deploy.yml
klutchell: We use the yocto build step to build extensions as well, we just target a different recipe.

.github/workflows/yocto-build-deploy.yml
klutchell: I've added a JSON schema, I'll go test it on the rpi repo

rcooke-warwick (approved):

klutchell:
Force-merging due to broken CI checks on generic-amd64.

PR #897, opened 2026-04-27
Add the rust crates to MIRRORS

klutchell:
Lately fetching the rust crates has been rate limited so let's make sure we
add our shared downloaded rust crates to MIRRORS in case we get fetching errors
from upstream.

See: https://balena.fibery.io/Security/Information_Security_and_Reliability_Incident/OS-pipeline-blocked-by-crates.io-rate-limiting-205


alexgg (approved):
lgtm

PR #900, opened 2026-05-05
build/barys: Add Wrynose build support

acostach:
When building with Poky Wrynose,
the oe-init-build-env script needs
to be called from the openembedded-core
layer, instead of poky.

Change-type: patch

### DIFF
--- a/build/barys
+++ b/build/barys
@@ -523,7 +523,15 @@ if [ -d "${TEMPLATECONF}/default" ]; then
 	TEMPLATECONF="${TEMPLATECONF}/default"
 fi
 
-source ${SCRIPTPATH}/../../layers/poky/oe-init-build-env ${SCRIPTPATH}/../../${BUILD_DIR}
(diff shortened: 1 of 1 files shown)


klutchell:
Sigh, we just pinned this last week via https://github.com/balena-os/balena-yocto-scripts/pull/899
```
4.120 Package docker.io is not available, but is referred to by another package.
4.120 This may mean that the package is missing, has been obsoleted, or
4.120 is only available from another source
4.120 
4.123 E: Version '27.*' for 'docker.io' was not found
```
So I guess we will have to find a new version to pin to that is still on the old API, or start using the Docker apt repos to get custom versions not included in Ubuntu repos.

We could pin to 20.x which should still be present and is less likely to be removed, but that is a bigger rollback. Until recently we were using v28 without issue as you can see on my PR.

Fixing in https://github.com/balena-os/balena-yocto-scripts/pull/901

alexgg (approved):
lgtm

PR #902, opened 2026-06-03
yocto-build-deploy: normalize placeholders, soft-skip schema validation

klutchell:
Two follow-ups to #895 before downstream adopters need them.

**Normalize composition placeholders to `__VAR__` form.** `$DEVICE_REPO_REV`
followed compose-spec's variable interpolation syntax (`$VAR` / `${VAR}`),
which would silently interpolate to empty if any future compose parser in
the deploy path enables it. `__DEVICE_REPO_REV__` is syntactically inert.

**Soft-skip the JSON Schema check when the bundled schema isn't in the checkout.**
Device repos haven't all bumped their `balena-yocto-scripts` submodule pointer
to a ref containing `schemas/hostapp-composition.json` yet. Skip with a
`::warning::` annotation; validation activates automatically once they bump.

Also adds a Placeholders section to the spec.

See: https://balena.fibery.io/Work/Improvement/Extensions-build-pipeline-changes-4332

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -406,7 +406,7 @@ jobs:
                 io.balena.image.store: 'root'
                 io.balena.image.class: 'hostapp'
                 io.balena.update.requires-reboot: '1'
-                io.balena.private.hostapp.board-rev: '$DEVICE_REPO_REV'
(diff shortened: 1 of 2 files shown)


rcooke-warwick (approved):

klutchell:
Force-merging to bypass the broken generic-amd64 CI job https://balena.fibery.io/Work/Improvement/Unblock-the-generic-amd64-pipeline-4158

PR #906, opened 2026-06-26
yocto-build-deploy: replace pattern based all history first parent match

alexgg:
Instead, resolve the PR commit by walking the pushed ref's own first-parent history.

Pattern search returns the youngest match reachable from ANY ref, so it can jump onto an unrelated branch (e.g. an ESR PR) whose commit is not even an ancestor of this tag.

Also,  match any `#<num>` instead of the literal "Merge pull request" prefix, so squash merged and custom titled merges
(like "Update GitHub Actions (#759)") are found, but not version-bump commits made by bots.

Change-type: patch

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -642,8 +642,13 @@ jobs:
         id: set-merge-commit
         if: ${{ github.event_name == 'push' }} # Only perform on push event - i.e a new version tag
         run: |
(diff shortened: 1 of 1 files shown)


rcooke-warwick:
This looks good to me, and I think we should merge it as until it is merged and deployed into the device repos, the finding of the correct merge commit will not work correctly, if I understand the situation right

rcooke-warwick (approved):

PR #907, opened 2026-07-02
One composition-driven publishing path for every built service

klutchell:
The hostapp published its release assets through two hardcoded glob lists and a
single `hostapp-artifacts` blob, while extensions flowed declaratively through the
composition. That asymmetry duplicated the contract logic, and once a device builds
more than one service their images and assets could stomp on each other.

This puts every built service, hostapp included, through the same composition-driven
flow, so each one's image and assets are namespaced and can't collide:

- Each service uploads one `image-<service>` artifact and one `release-assets-<service>`
  artifact. The deploy job downloads them per service, loads each image and matches it
  back to its service, then attaches the hostapp's assets with flat keys and each
  extension's under `service/<name>/`.
- The hostapp is now just another composition service. Its image follows the same
  `<recipe>-<machine>.docker` name as extensions, defaulting the recipe to `balena-image`
  since the hostapp declares none, and its release assets come from `x-build.assets`.
  `balena_stage_release_assets` stages the few non-native files into the yocto deploy
  dir and everything uploads from there.
- Dropped the `hostapp-artifacts` blob, the `S3_DEPLOY_PATH` staging area, and
  `balena_deploy_artifacts`.

Also folded in while the paths were being reworked:

- All artifacts move through `lzma-artifact-action`, one composite action for LZMA
  compression plus optional whole-archive AES-256, replacing `upload/download-artifact`
  and the hand-rolled openssl encrypt/decrypt steps.
- Raw-image DEFLATE compression is a stdlib `prepare-image.py` now, so no helper image.
- A Verify step checks the contractually required release assets are attached before
  finalizing. Releases deploy as a draft and finalize last, so devices never see an
  incomplete release. See [related project](https://balena.fibery.io/Work/Project/Ensure-OS-build-pipeline-has-expected-artifacts-before-uploading-them-as-release-assets-2658).
- Removed archive device-type support.

Tested on staging: rpi5 ran the full pipeline through hostapp deploy and finalize;
generic-amd64 exercised the encrypted upload and download path.

Change-type: minor
See: https://balena.fibery.io/Work/Project/hostapps-One-composition-driven-publishing-path-for-every-built-service-2773


.github/workflows/yocto-build-deploy.yml
```diff
@@ -1934,58 +1924,202 @@ jobs:
...
+      # A finalized release missing these is broken, so fail loudly here instead.
+      # Contract: docs/specs/hostapp-release-assets.md
+      - name: Verify required release assets
+        if: needs.balena-lib.outputs.should_finalize == 'true'
+        env:
+          BALENAOS_TOKEN: ${{ secrets.BALENA_API_DEPLOY_KEY }}
+          API_ENV: ${{ env.BALENARC_BALENA_URL }}
+          RELEASE_ID: ${{ steps.deploy-hostapp.outputs.release_id }}
+          # release_asset is a v7-only resource (v6 has no translation for it).
+          CURL: "curl --silent --retry 10 --location --compressed"
```
alexgg: Maybe better check for failure here than deal with an unspecified error later on.

.github/workflows/yocto-build-deploy.yml
```diff
@@ -2029,34 +2167,22 @@ jobs:
 
     steps:
 
-      - name: Download artifacts
-        uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7.0.0
+      # S3 needs the composition's release assets (deflates, manifests, image*.json, device-type.json,
+      # licenses, cyclonedx, ...) but NOT the deploy docker image. release-assets-hostapp is exactly
```
alexgg, line 2182: Are we dropping the hostapp docker image from S3 then as it's pullable from the registry? I'll miss the convenience of just downloading it from S3/build jobs though.

alexgg (commented):
lgtm, couple of nitpicks and of it goes.

.github/workflows/yocto-build-deploy.yml
klutchell: @alexgg we are dropping the docker image from S3, but it will still be available as a GitHub artifact.

Soon all S3 assets will be removed anyway once we finish deprecating support for the legacy S3 upload path.

.github/workflows/yocto-build-deploy.yml
klutchell: Good call, added `--fail-with-body` and error handling

alexgg (approved):
lgtm

PR #917, opened 2026-08-11
Allow deploying with a balena CLI release channel

cywang117:
Add a `cli-channel` input that swaps the pinned CLI version for a release channel build in the hostApp deploy, so an unreleased CLI can be tested. The action treats `cli-version` and `cli-channel` as mutually exclusive, so the version is passed only when no channel is requested.

Bump setup-balena-action to v0.1.0, which adds `cli-channel`.

Change-type: minor

### DIFF
--- a/.github/workflows/yocto-build-deploy.yml
+++ b/.github/workflows/yocto-build-deploy.yml
@@ -1684,8 +1684,14 @@ jobs:
           # renovate: datasource=github-releases depName=balena-io/balena-cli
           BALENA_CLI_VERSION: v25.1.10
         with:
-          # balena CLI version to install
-          cli-version: ${{ env.BALENA_CLI_VERSION }}
+          # A channel is named after a balena CLI pull request branch and moves as
+          # that branch gains commits, so this installs unreviewed code into the
+          # hostApp deploy job. Drop it and restore cli-version below once the CLI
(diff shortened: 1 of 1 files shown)


.github/workflows/yocto-build-deploy.yml
```diff
@@ -71,6 +71,15 @@ on:
...
         type: string
+      # A channel is named after a balena CLI pull request branch and moves as that
+      # branch gains commits, so this installs unreviewed code into the hostApp
+      # deploy job. Set it to test an unreleased CLI build, and leave it unset
+      # otherwise to keep the pinned release.
+      cli-channel:
+        description: balena CLI release channel to use for the hostApp deploy, instead of the pinned version
+        required: false
+        type: string
+        default: ''
```
klutchell: I don't think we need an input, we will probably just hardcode the channel in the workflow for now. Unless we see a need to regularly re-pin the CLI to different release channels on a per-device repo basis, I think this will just add complexity and inputs are difficult to deprecate once calling workflows are using them.

cywang117: Thanks for looking! I hard-coded the cli-channel to `os-profiles`.

klutchell (approved):

PR #925, opened 2026-09-02
yocto-build-deploy: Use authenticated user to update submodules

klutchell:
Avoid GitHub rate-limiting of self-hosted runners by using an authenticated user when updating submodules.

Change-type: patch


mtoman (approved):
lgtm

